PR #188, opened 2015-01-15
close post connections on completion

mtmorgan:
PUT(..., body=upload_file()) leaves an open connection, so PUTing many files exhausts the pool of available connections (cf., comment in original code). This pull request closes the connection when its input is complete; the connection might remain open / in use if an error occurs.


hadley:
Thanks! Would you mind adding a bullet to news too please? 

It might also be a good idea to wrap that function in `safe_callback()` - I think currently RCurl will crash if you abort during a long upload?

PR #192, opened 2015-02-12
make_request is using partial matching

daattali:
disclaimer: I haven't tested the proposed change.  I was getting an warning when using `devtools::install_github` that propagated from doing a partial match here of `status` to `status_code`


hadley:
Thanks!

PR #202, opened 2015-03-03
Fix service account token signature.

siddharthab:
We never did end to end testing of service account tokens. I am surprised no one complained. :)


hadley:
Oops! I need to add some tests for this :/

PR #206, opened 2015-03-08
explicit setNames

robertzk:
This has been giving me some trouble when using httr from within other package namespaces.


peterhurford:
:+1:

hadley:
Thanks!

Issue #210, opened 2015-03-18
oauth listener is hard-wired to listen only on `127.0.0.1`


cboettig:
Finally figured out what my problem was in issue #191 (which actually isn't at all related to `use_obb` so I've just closed that issue).  

Your OAuth listener function is hardwired to use `127.0.0.1`.  This is incompatible with how Docker containers work, which expect the host to be `0.0.0.0` so that we can link the 1410 port to the host machine.  (e.g. same thing happens if you run Jekyll inside of a container, you just have to set host to 0.0.0.0).  Most software lets us do this (e.g. no problem in servr etc running in a container), but httr has this hard wired to the `localhost` value of `127.0.0.1`, https://github.com/hadley/httr/blob/e746e973e18504996f3d4916c9f2bba334a85ac8/R/oauth-listener.r#L48.  Think you can change this?

Not sure what the best way to expose an option where we can set this would be (e.g. would it make sense to just add `host = 127.0.0.1` as an optional argument to `oauth1.0_token()` function (and any other functions that lead to a call to `oauth_listener()`)?

Thanks for considering; it would be just awesome to have this fixed. I'd work up a pull request but thought I better check in with you first since I'm not sure how best to get the user interface right here.

yutannihilation:
I've faced the same problem with RStudio Server running inside a vagrant VM. It's awesome if I can change the `host` IP!

hadley:
If I did change this, what would you set the host address to in the individual app settings? If it's not 127.0.0.1, the response won't get re-directed to the right place.

cboettig:
@hadley I've put in a pull request with this feature in PR #211; it sets 127.0.0.1 as the default but just provides a way for a user to change it.

PR #211, opened 2015-03-18
add support for variable host and port as suggested in #210

cboettig:
A heck, here's a PR to implement this anyway, that just makes `host` and `port` into optional function arguments. Let me know if you'd rather it were done differently.  Many thanks

### DIFF
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -7,8 +7,7 @@
 #'    \code{\link{oauth_app}}
 #' @param permission optional, a string of permissions to ask for.
 #' @param is_interactive Is the current environment interactive?
-#' @param host ip address for the listener
-#' @param port for the listener
+#' @inheritParams oauth_listener 
 #' @export
 #' @keywords internal
 init_oauth1.0 <- function(endpoint, app, permission = NULL,
@@ -53,6 +52,7 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
 #'     Otherwise, provide a URL to the user and prompt for a validation
 #'     code. Defaults to the of the \code{"httr_oob_default"} default,
 #'     or \code{TRUE} if \code{httpuv} is not installed.
+#' @inheritParams oauth_listener
 #' @export
 #' @keywords internal
 init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -7,10 +7,13 @@
 #'    \code{\link{oauth_app}}
 #' @param permission optional, a string of permissions to ask for.
 #' @param is_interactive Is the current environment interactive?
+#' @param host ip address for the listener
+#' @param port for the listener
 #' @export
 #' @keywords internal
 init_oauth1.0 <- function(endpoint, app, permission = NULL,
-                          is_interactive = interactive()) {
+                          is_interactive = interactive(),
+                          host = "127.0.0.1", port = 1410) {
 
   oauth_sig <- function(url, method, token = NULL, token_secret = NULL, ...) {
     oauth_header(oauth_signature(url, method, app, token, token_secret, ...,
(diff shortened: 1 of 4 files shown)


cboettig:
@hadley ping. I know httr is still in the queue behind lots of other good stuff; but this is a quick fix to an issue that bites me regularly. (I'd just use my branch but seems weird to tell users to do that)

hadley:
I'll add httr to my list of packages that need updates (it's currently behind devtools and tidyr)

hadley:
Could you add a bullet point to NEWS please?

R/oauth-init.R
```diff
@@ -7,10 +7,13 @@
 #'    \code{\link{oauth_app}}
 #' @param permission optional, a string of permissions to ask for.
 #' @param is_interactive Is the current environment interactive?
+#' @param host ip address for the listener
```
hadley: `@inheritParams ouath_listener` would be a bit DRYer

cboettig: Thanks for these, will do

hadley:
Thanks!

cboettig:
@hadley Now I realize that it probably would have been better to set the default of the function argument to something like `getOption("httr_localhost")` since most third-party R packages define their own auth wrappers around httr and aren't exposing this option either.  

To my thinking, a better implementation would actually check the environmental variable for the default, since I could then set something like `ENV LOCALHOST 0.0.0.0` inside the container and everything would work with no additional intervention.  But I'm not sure if calling `Sys.getenv("LOCALHOST")` is a good idea.  Thoughts?

cboettig:
@hadley More problematically, I'm worried that this PR might have introduced a bug to oauth as well.  I currently get the following error with traceback:

```
 httr::oauth2.0_token(dropbox, dropbox_app)
Error in makeTcpServer(host, port, appWrapper$onHeaders, appWrapper$onBodyData,  : 
  object 'host' not found

Enter a frame number, or 0 to exit   

1: httr::oauth2.0_token(dropbox, dropbox_app, host = "0.0.0.0")
2: Token2.0$new(app = app, endpoint = endpoint, params = params, cache_path = 
3: public_bind_env$initialize(...)
4: self$init_credentials()
5: init_oauth2.0(self$endpoint, self$app, scope = self$params$scope, type = se
6: oauth_listener(authorize_url, is_interactive, host, port)
7: httpuv::startServer(host, port, list(call = listen))
8: makeTcpServer(host, port, appWrapper$onHeaders, appWrapper$onBodyData, appW
```

If I go into the environment at `6`, `oauth_listener` it says:

```
Selection: 6
Browse[4]> ls()
[1] "host"           "info"           "is_interactive" "listen"        
[5] "port"           "request_url"   
Browse[4]> host
Error during wrapup: promise already under evaluation: recursive default argument reference or earlier problems?
```

even though at `1` the value `host` is correctly identified.  I'm not sure if something has changed since the PR, but at that time I could confirm that this set the host correctly without error.  Not sure what's up.      maybe I should re-open or open a new issue?

hadley:
An env var seems ok to me, but it should be something like `HTTR_LOCALHOST`. I'd recommend making a little function to wrap it up.

I can't replicate the bug you're seeing

cboettig:
@hadley It's definitely a bug.  Looks like I should have edited the R6 Token class to include the `host` variable and define `self$host <- host`, e.g. https://github.com/hadley/httr/blob/master/R/oauth-token.r#L40-L45  

My apologies, I completely overlooked what was going on in the R6 classes when working out this pull request but @richfitz set me straight.  

```
install_github("cboettig/rdrop2")
rdrop2::drop_auth(host="0.0.0.0")
```

should replicate this issue(?)

It looks like I need to modify https://github.com/cboettig/httr/blob/master/R/oauth-token.r#L40-73 to include `host` and `self` as part of the definition, but I'm hesitant to go hacking fundamental class structures without really knowing what I'm doing...

hadley:
I think you'll need to add a listener field that stores the host and port (and anything else that might come up in the future)

PR #212, opened 2015-03-23
Use explicit rather than partial matching

jimhester:
https://github.com/hadley/httr/commit/e415faf42f1260b59bbe30e8648d4999fe83cbee fixed the partial match in the httr code, but there were a few more places in the tests that were also using partial matching.

found by running the tests with

```r
options(warnPartialMatchAttr = TRUE, warnPartialMatchDollar = TRUE,
        warn = 2)
```


hadley:
Thanks!

PR #213, opened 2015-03-26
Added check for existing cainfo config setting

jrowen:
The current behavior on Windows is to set always reset `cainfo` to the default package file, regardless of whether the user has already specified a value for the parameter.  This update adds a check to verify that `cainfo` has not already been set before setting to the default.

This is a re-submission of #199.

### DIFF
--- a/R/config.r
+++ b/R/config.r
@@ -196,7 +196,8 @@ default_config <- function() {
     user_agent(default_ua()),
     add_headers(Accept = "application/json, text/xml, application/xml, */*"),
     write_memory(),
-    if (.Platform$OS.type == "windows") config(cainfo = cert),
+    if (.Platform$OS.type == "windows"  &&
+      !("cainfo" %in% names(getOption("httr_config")))) config(cainfo = cert),
     getOption("httr_config")
   )
 }


hadley:
Are you sure this is actually a problem? `getOption("httr_config")` comes after the default setting so it will override.

jrowen:
Prior to this change, I see the following output on my Windows machine.
```
> library(httr)
> reset_config()
> set_config(httr::config(cainfo="mycert.cer"))
> httr:::default_config()
Config: 
List of 8
 $ followlocation:TRUE
 $ maxredirs     :10
 $ encoding      :"gzip"
 $ useragent     :"curl/7.39.0 Rcurl/1.95.4.5 httr/0.6.1"
 $ writer        :List of 1
  ..$ buffer: NULL
  ..- attr(*, "class")="write_memory" "write_function"
 $ cainfo        :"C:/Users/jrowen/Documents/R/win-library/3.1/httr/cacert.pem"
 $ cainfo        :"mycert.cer"
 $ httpheader    :"application/json, text/xml, application/xml, */*"
  ..- attr(*, "names")="Accept"
```
The first `cainfo` option appears to be the one used.  After applying this code update, as expected, I only see the second `cainfo` option.

hadley:
Hmmmmm, it does seem like the first option is used instead of the last:

```R
GET("http://httpbin.org/user-agent", config(useragent = "a", useragent = "b"))
```

I'd rather fix that general problem than the specific cainfo prob.

PR #217, opened 2015-04-13
Update oauth2-github.r

johnzeng2010:
This added line is necessary for me to go through.


hadley:
It shouldn't be. As long as you have it installed, the existing code should work.

PR #226, opened 2015-05-09
Handle doc

antoine-lizee:
As discussed in [issue#219](https://github.com/hadley/httr/issues/219#issuecomment-100223384).

PR #227, opened 2015-05-09
Family inclusion

antoine-lizee:
A small fix in the doc to see all the different http methods in the corresponding "see also" sections.


antoine-lizee:
I would be happy to recompile the help files on the top of PR #226.

hadley:
Gah, wrong one.

PR #232, opened 2015-05-15
rm trailing slash in url (hostname) as redundant.

nicolasCoutin:
https://api.github.com/ --> https://api.github.com

The `hostname` of `url` _itself_ doesn't/?shouldn't actually contain the trailing slash.

### Eg:
```r
library(httr)
slashy <- modify_url("https://api.github.com/",path = 'my-path')
no_slashy <- modify_url("https://api.github.com",path = 'my-path')
identical(slashy, no_slashy)
#> [1] TRUE
```

Or if you should use the `slashy` version I would love an explanation (sorry if it's very obvious).


hadley:
Thanks!

PR #233, opened 2015-05-17
tumblr

isomorphisms:
https://www.tumblr.com/docs/en/api/v2#auth


hadley:
Could you please match the style of the other entries, and add a bullet point to NEWS?

hadley:
It would also be useful to add a full demo to the demos directory

isomorphisms:
Sure. Should I also change twitter, vimeo, and yahoo to look like google and facebook?

hadley:
Thanks!

PR #234, opened 2015-05-17
BROWSE didn't work for me..

isomorphisms:
This may be too messy to be accepted, but I had to `fixInNamespace` and add the URL so I could manually browse to it.

Maybe everyone else knows to do this as well, but I had to spend a bit of time on it.

### DIFF
--- a/R/oauth-listener.r
+++ b/R/oauth-listener.r
@@ -50,6 +50,7 @@ oauth_listener <- function(request_url, is_interactive = interactive(), host = "
   server <- httpuv::startServer(host, port, list(call = listen))
   on.exit(httpuv::stopServer(server))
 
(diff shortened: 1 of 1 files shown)


hadley:
What is the problem?

isomorphisms:
Maybe I should just have filed this as an issue. `BROWSE` does not open up a browser for me on Ubuntu 12.04 LTS, so I needed to print the check-in URL to the screen so I could manually navigate there. That required digging into the code to find out how to get the function to print. Might save someone else with a similar problem a couple minutes, although I can understand how you might think it's uglier this way.

hadley:
You probably just need to set `option(browser)` to something useful.

PR #235, opened 2015-05-20
Work around listener errors introduced in #211

cboettig:
@hadley I've taken the short-cut of just modifying the listener method to read the environmental variables directly, and removed the option to pass down explicit host and port in `oauth_init` and `oauth_token` functions.

I did try modifying the R6 initialization structures to include a `listener` object that would store this, but am not very solid on R6 methods and their inheritance behavior; or how this would interact with caching.  Making so many changes to internal plumbing just to add support for a feature most people would only configure with environmental variables, if at all, seemed unwise, so I opted for this cop-out.  

Sincere apologies for the incomplete implementation in #211 and also for the imperfect git log history here...


R/oauth-listener.r
```diff
@@ -77,3 +77,14 @@ oauth_listener <- function(request_url, is_interactive = interactive(), host = "
 oauth_callback <- function() {
   "http://localhost:1410/"
 }
+
+listener_endpoint <- function ()
```
hadley: Can you please follow house style? (e.g. http://r-pkgs.had.co.nz/r.html#style)

R/oauth-listener.r
```diff
@@ -77,3 +77,14 @@ oauth_listener <- function(request_url, is_interactive = interactive(), host = "
...
+{
+  host <- Sys.getenv("HTTR_LOCALHOST")
```
hadley: Maybe `Sys.getenv("HTTR_LOCALHOST", "127.0.0.1")` ?

daattali:
super useful, thanks Carl

hadley:
Thanks! BTW I only get notifications on comments, not additional commits, so when you're done with requested changes it's helpful to comment with PTAL or similar.

PR #237, opened 2015-05-22
Avoid having to set ip address twice for non-docker server

cboettig:
As @daattali pointed out to me, we should avoid having remote server users having to set the IP twice (in both `HTTR_LOCALHOST` and `HTTR_SERVER`, while still covering the docker case where the internal IP set up by httpuv differs from the external ip (on docker the internal one is always 0.0.0.0).  

Now any remote sever user only needs to set the IP in `HTTR_SERVER`.  The rocker images will set `HTTR_LOCALHOST` by default, so docker users won't need to realize this is different.  Non-docker users won't have `HTTR_LOCALHOST` set at all, and so the code will default to the same value as `HTTR_SERVER` now, instead of `localhost`, so the whole thing is a bit more streamlined.  

@hadley Sorry for the multiple commits & PRs to get this right!

### DIFF
--- a/R/oauth-listener.r
+++ b/R/oauth-listener.r
@@ -76,12 +76,12 @@ oauth_listener <- function(request_url, is_interactive = interactive()) {
 #' @export
 oauth_callback <- function() {
 	paste0("http://", 
-               Sys.getenv("HTTR_SERVER", "localhost"),
+               Sys.getenv("HTTR_SERVER", "127.0.0.1"),
                ":",
                Sys.getenv("HTTR_SERVER_PORT", "1410"))
 }
 
 listener_endpoint <- function() {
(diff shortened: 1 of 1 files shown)


hadley:
I'm not sure that will work in general - IIRC a number of OAuth implementations don't allow the use of localhost as the callback url

cboettig:
Right, I've used `127.0.0.1` instead of `localhost` in both places, https://github.com/hadley/httr/pull/237/files#diff-57e3b027bcc9892890b953444e44fa22R85
instead of ever calling `localhost` explicitly. Do you think that's likely to be a problem?

hadley:
Yes, I think it is. I'd rather put off this change until we figure out how to unit test OAuth, otherwise I think the risk of breakage is high.

PR #254, opened 2015-06-29
Add progress(type="both")

jeroenooms:
Test:

```
POST("https://httpbin.org/post", progress("both"), body = as.character(rnorm(1e6)))
```

### DIFF
--- a/R/progress.R
+++ b/R/progress.R
@@ -10,7 +10,7 @@
 #' # Otherwise you get the number of bytes downloaded:
 #' x <- GET("http://httpbin.org/drip?numbytes=4000&duration=3", progress())
 #' }
-progress <- function(type = c("down", "up")) {
+progress <- function(type = c("down", "up", "both")) {
   type <- match.arg(type)
 
   request(options = list(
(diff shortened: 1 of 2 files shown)


R/progress.R
```diff
@@ -24,7 +24,13 @@ progress_bar <- function(type) {
   first <- TRUE
 
   show_progress <- function(down, up) {
-    if (type == "down") {
+    if (type == "both") {
+      total <- 1
```
hadley: Why not make `total <- down[[1]] + up[[1]]` ?

R/progress.R
```diff
@@ -10,7 +10,7 @@
 #' # Otherwise you get the number of bytes downloaded:
 #' x <- GET("http://httpbin.org/drip?numbytes=4000&duration=3", progress())
 #' }
-progress <- function(type = c("down", "up")) {
+progress <- function(type = c("down", "up", "both")) {
```
hadley: Maybe "both" should be the default?

R/progress.R
jeroenooms: down[[1]] is defined by the `content-length` response header so it is unknown until the upload has completed (I think)

hadley: Oh hmmm. Maybe it would be better to do two separate progress bars? It seems like usually either the upload or download is long, not both

jeroenooms: Yes maybe.

hadley:
I'd rather both create two separate progress bars - I think combining them is likely to be confusing

PR #260, opened 2015-07-01
Add support for OAuth 2.0 implicit grant flow

kippandrew:
The implicit flow (outlined https://tools.ietf.org/html/rfc6749#section-1.3.2) is oauth2 grant that is used for public (e.g. non-confidential clients). This type of client is typically browser based (javascript) applications where the client can not maintain confidentiality. However, this flow may also be useful for applications such as publicly disturbed R packages. 

The implicit flow is less secure then the authorization code flow (see http://tools.ietf.org/html/rfc6749#section-10.16). Also, not all providers implement this type of grant. One important note is that implicit flows do not issue refresh tokens.

### DIFF
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -51,11 +51,13 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
 #'     code. Defaults to the of the \code{"httr_oob_default"} default,
 #'     or \code{TRUE} if \code{httpuv} is not installed.
 #' @param is_interactive Is the current environment interactive?
+#' @param use_implicit
 #' @export
 #' @keywords internal
 init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
                           use_oob = getOption("httr_oob_default"),
-                          is_interactive = interactive()) {
+                          is_interactive = interactive(),
+                          use_implicit = FALSE) {
   if (!use_oob && !is_installed("httpuv")) {
     message("httpuv not installed, defaulting to out-of-band authentication")
     use_oob <- TRUE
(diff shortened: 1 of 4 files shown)


hadley:
Do you have any thoughts on how it might be possible to test this?

kippandrew:
A few approaches:

To fully unittest oauth2 clients in the past, I've used stub http servers that pretend to be the oauth2 provider. The stub server can return various responses to mimic what a real oauth2 auth server would do. This might a bit overkill, but it could be useful for testing all the other flows. I could look at some type of test harness for testthat/httuv.

If you're looking for a demo like we have for the other providers, I think Yahoo supports implicit type grant. I can get that working also. 

This is flow was tested against shinyapps.io new oauth provider which is not yet in production.

hadley:
* Yeah, a demo would be good. Is it possible to use shinyapps now?
* Can you please also add a bullet point to NEWS?
* It would be useful to include that link to the spec in the docs.

kippandrew:
I'll work something up here. The shinyapps.io OAuth2 Authorization server is live, so we can test against that. Long term plans is we like to have the `shinyapps`/`rsconnect` packages use this mechanism for authentication.

hadley:
It's also worth considering if this should be split out into `init_oauth2.0_implicit()`

kippandrew:
That might be a cleaner approach. I'll take a look.

hadley:
Do you still want to get this in? I'm aiming to release early next week, but I can hold back if you think this is important.

kippandrew:
No need to wait on this. Not sure I have the bandwidth to look at this at
the moment.

hadley:
No problems. I'll close for now.

mariafb:
Hi Guys. Were you finally able to implement httr for Oauth2.0 implicit grant flow? If not, do you know any other source that could help me. 
Many Thanks !

PR #264, opened 2015-07-04
Small cleanup.

craigcitro:
After the fix in jeroenooms/curl#32, we might as well drop these
`as.character` calls, since they're now unnecessary.

PTAL @hadley


hadley:
Thanks!

PR #274, opened 2015-08-25
Follow de facto GITHUB_PAT convention

jennybc:
See this thread https://github.com/ropensci/git2r/issues/170#issuecomment-134704010. (I should have done this from start.)


hadley:
Thanks!

PR #280, opened 2015-10-02
build_url collapse path

artemklevtsov:
Collapse `path` argument if it length more than 1.

### DIFF
--- a/R/url.r
+++ b/R/url.r
@@ -107,6 +107,10 @@ build_url <- function(url) {
     port <- NULL
   }
 
+  if (length(url$path) > 1) {
+    url$path <- paste(url$path, collapse = "/")
+  }
+
   path <- url$path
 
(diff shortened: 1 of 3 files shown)


R/url.r
```diff
@@ -107,6 +107,10 @@ build_url <- function(url) {
     port <- NULL
   }
 
+  if (length(url$path) > 1) {
+    url$path <- paste(url$path, collapse = "/")
+  }
+
   path <- url$path
```
hadley: I think you should just skip the if - there's no harm to pasting if length is 1. Then you can combine with the next line.

artemklevtsov: Ok, done.

hadley:
And can you please add a bullet point to NEWS?

hadley:
Thanks!

PR #285, opened 2015-10-26
Remove the stringr dependency

jimhester:
Stringr is a very heavy dependency (due to the stringi dependency) which means any package which uses httr has to compile this heavy dependency. This makes packages (such as devtools) take a large amount of time to build (such as on Travis). While I generally feel one should not fear heavy dependencies for most packages, because httr is a widely used foundation package having such a large dependency is not ideal.

This pull request removes this dependency by providing drop in R only replacements for the equivalent stringr functions written using base functions.

They don't provide the full functionality of the stringr functions, but they provide what is needed by the usage in httr.

All the tests and examples pass on my machine.


hadley:
Thanks - I agree it's good to minimise dependencies in these infrastructure packages.

(Ooops, forgot to also ask for a bullet point in NEWS)

jimhester:
I'll open a new PR with the NEWS change

jeroenooms:
:thumbsup: for lite packages.

PR #287, opened 2015-10-26
Do not partial match header

jimhester:
Noticed these when I was running tests.

Can I send a PR to @hadley's personal .Rprofile

```r
options(warnPartialMatchDollar = TRUE)
```

:smiley:


hadley:
It would be useful to add those options automatically in devtools when checking.

Issue #291, opened 2015-11-14
request_prepare kills POST calls with token


ReportMort:
I am receiving the error `Error: is.request(y) is not TRUE` while trying to perform a POST call with an oauth2.0_token. My POST call is very similar to the demos:

```
POST(url=url, config=config(token=my_token), body=this_request_body)
```

I've noticed that `POST()` calls `request_perform()` which calls `request_prepare()`, which checks if there is a token and signs the request. The problem occurs at https://github.com/hadley/httr/blob/f7593b7d5b3f9396469d1220dd4b38206fbb7d2e/R/request.R#L107 when using the `c()` operator to push the request and signed token together. This operation forces a check that `token$sign(req$method, req$url)` is a valid request object, but it's not, so the code bombs. `token$sign(req$method, req$url)` actually returns a list of length 2 containing a url string and a config (which is a valid request by itself). The url seems redundant and I think it will be inherited from the actual `req` object when the `%||%` function runs, so it can really be ignored.

Everything works fine when I peel off the config portion of the `token$sign` operation:
```
req <- c(req, token$sign(req$method, req$url)$config)
```

hadley:
Can you provide a reproducible example? The demos work for me.

PR #292, opened 2015-11-14
Fix request_prepare and trimws dependency

ReportMort:
Update request_prepare so that when the token is signed, then it returns another request object in order to successfully be combined back to the original request. This is in reference to Issue: https://github.com/hadley/httr/issues/291

### DIFF
--- a/R/request.R
+++ b/R/request.R
@@ -104,7 +104,7 @@ request_prepare <- function(req) {
   # Sign request, if needed
   token <- req$auth_token
   if (!is.null(token))
-    req <- c(req, token$sign(req$method, req$url))
+    req <- c(req, token$sign(req$method, req$url)$config)
 
   req
 }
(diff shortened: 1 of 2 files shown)


ReportMort:
Also, see Issue: https://github.com/hadley/httr/issues/293. This pull request adds the code for `trimws` directly so that a dependency on R 3.3.0 isn't required.

hadley:
I'd rather include just something simple like:

```R
str_trim <- function(x) {
  gsub("(^\\s+)|(\\s+$)", "", x
}
```

so we don't have to add R-core to the authors, and change the LICENSE etc.

ReportMort:
Sounds good. I switched the code to use your proposed version of `str_trim`, but I may just close the PR entirely depending on your response to my first issue (https://github.com/hadley/httr/issues/291).

hadley:
Thanks - I'd rather tackle these two issues separately.

PR #298, opened 2015-12-17
Switch to openssl for hmac, hashing, signatures, base64, etc

jeroenooms:


R/oauth-server-side.R
```diff
@@ -27,13 +27,9 @@ init_oauth_service_account <- function(endpoint, secrets, scope = NULL) {
 #' jwt_signature(cred, "https://www.googleapis.com/auth/userinfo.profile")
 #' }
 jwt_signature <- function(credentials, scope, duration = 60 * 60) {
-  if (!requireNamespace("PKI", quietly = TRUE)) {
-    stop("PKI package required for JWT signing", call. = FALSE)
+  if (!requireNamespace("openssl", quietly = TRUE)) {
```
hadley: You can drop this check now, since openssl is in imports

R/hmac.r
```diff
@@ -7,7 +7,10 @@
 #' @keywords internal
 #' @export
 hmac_sha1 <- function(key, string) {
-  hash <- digest::hmac(key, string, "sha1", raw = TRUE)
-  base64enc::base64encode(hash)
+  if(is.character(string))
```
hadley: Can you put a space after `if` please

hadley:
Looks good - my main concern is that there are few (no?) unit tests for these features.  Any thoughts on some simple tests that might be effective at catching regressions?

jeroenooms:
I think most is covered by the existing oauth unit tests, except for `jwt_signature`. There is a `\dontrun()` [example](https://github.com/hadley/httr/blob/master/R/oauth-server-side.R#L25-L28) for `jwt_signature` but it requires a local auth file. If this example works, we could store that file encrypted on travis.

hadley:
Ok, thanks!

I think @craigcitro is thinking about the `jwt_signature` type stuff and will hopefully have a unit-testable solution for us soon :)

PR #301, opened 2015-12-18
Azure Active Directory endpoint access

cornf4ke:
Dear Hadley,

I created a fork of httr as I noticed that there is not option to pass a resource URI to oauth2.0_token as required by the access process of Azure AD. I made the package work with Azure. The implementation as an additional parameter might not be desirable by you but I couldn't judge whether it would be .

The changes I made are:
1. As I am currently working on windows and have default enabled windows style checkouts, the test suite didn't pass after checkcout. I fixed this by always cleaning the eol characters before comparing. strings. (I just realized it might be strictly necessary only once.)
2. I added the additional parameter resource_id to the oauth2.0_token token request chain.
3. I added the common endpoint definition for Azure AD.
4. To make the package pass tests, I ran roxygen2 on it. This is the last commit.

Best regards,
Daniel


hadley:
Could you please merge/rebase? There seem to be a lot of spurious diffs

cornf4ke:
Sorry for not responding any quicker. Christmas and moving to a new place took all of my attention in between. I fetched your updated master and merged it into my branch. Tests ran without problems on my machine.

I am currently using my code to fix the additional argument problem for azure but was wondering  whether it might be more appropriate to allow for a list argument or "..." of additional named parameters and then c() the parameters into a single list while forming the request. I didn't do this because I don't know that much about oauth2 interface stability and possible future interface changes.

Best regards,
Daniel

hadley:
I'm not opposed to extending httr's oauth support to work with more sites, but there are a lot of minor variations seem to be only used by a single site.  I'm more likely to accept the change if you can point to where in the [OAuth2 RFC](https://tools.ietf.org/html/rfc6749) `resource_id` is mentioned (and how it's different to the usual `scope`)

cornf4ke:
I share your critisicm and would prefer if microsoft had chosen to just pass their specification of the requested resource via the scope parameter. For reasons which are not entirely clear to me, they chose to extend the protocol by a non-standard parameter "resource". The references which made me implement the optional parameter are:
* http://stackoverflow.com/questions/23921884/what-is-the-resource-parameter-in-windows-azure-ad-tenant-application-oauth-2-0
* https://books.google.de/books?id=5-U3CwAAQBAJ&pg=PT192&lpg=PT192&dq=oauth2+resource+parameter+azure+ad&source=bl&ots=aqrKt2trLS&sig=L5h0ZihauqQGPhYpTfCFhuhBKrg&hl=de&sa=X&ved=0ahUKEwj7kZaWupDKAhUo8XIKHaKlBsAQ6AEILTAC
The msdn documentation examples also use this parameter in their authentication procedure with no alternative.

I would change the implementation to one which passes an optional argument list "nonstandard_parameters" as this might be somewhat clearer to the user and does not require separate if you are willing to include it into the official package. I would be glad to be able to use the official resource as I might need to deploy some code on client sites in the MS Windows world in the near future.

hadley:
Yeah, I'd prefer something like "user_params" which took a named list.

cornf4ke:
Ok. I will change this accordingly and then open a new pull request on a cleaner branch.

Issue #275, opened 2015-08-28
TokenServiceAccount not working - does not get passed into config() with GET


makanig:
I filed an earlier issue on oauth_service_token.
After fixing that, I found this other problem.
For this one, I don't quite know the best way to fix this one, and as a workaround I used the query list for the GET() to include the service_token.

Here's the issue:
```
googleToken <- # I got the token ..TokenServiceAccount$new
> req <- GET(surl, config(token=googleToken)) # this seems to be the documented way
Error: is.request(y) is not TRUE
> req <- GET(surl, token=googleToken)  # this does not pass the token into the GET 
> content(req)
...
$error$errors[[1]]$message
[1] "Login Required"
```

hadley:
Could you please provide a reproducible exampe?

MarkEdmondson1234:
I have one here which relates to two linked issues I think when working with service tokens, I think the below should work but perhaps I'm not understanding how service tokens should work:

You need to put in your own Google project secrets, JSON files, tokens etc.:

```r
## example with URL shortner API
SCOPE <- "https://www.googleapis.com/auth/urlshortener"
JSON_SECRET <- "yourJSONsecretfile.json" ## for service auth
PROJECT_WEBAPP_KEY <- "xxxxx" ## for OAuth2 auth
PROJECT_WEBAPP_SECRET <- "dddddd" ## for OAuth2 auth

oauth2_token <- function(){
  endpoint <- httr::oauth_endpoints("google")
  
  app <- httr::oauth_app("google", 
                         key = PROJECT_WEBAPP_KEY, 
                         secret = PROJECT_WEBAPP_SECRET)
  
  google_token <- httr::oauth2.0_token(endpoint = endpoint, 
                                       app = app,
                                       scope = SCOPE, 
                                       cache = ".httr-oauth")
  
}

service_token <- function(secrets){
  endpoint <- httr::oauth_endpoints("google")
  scope <- SCOPE
  secrets <- jsonlite::fromJSON(secrets)
  
  ## multiple scopes only work if space separated
  scope <- paste(scope, collapse=" ")
  
  ## client ID etc. not needed as in JSON secrts
  google_token <- httr::oauth_service_token(endpoint, secrets, scope)
}

#### Examples

## OAuth2 - works
o2_token <- oauth2_token() ## go through authentication flow

arg_list <- list(url = "https://www.googleapis.com/urlshortener/v1/url?shortUrl=https://goo.gl/cFVQbk", 
[...]

PR #304, opened 2015-12-21
fix for token service account

nathangoulding:
Fixes https://github.com/hadley/httr/issues/275


nathangoulding:
I don't believe this is the correct fix.

PR #305, opened 2015-12-21
expose status code for external library future robustness

nathangoulding:
This exposes the actual HTTP status code so that external libraries (like bigrquery) can rely on that instead of parsing text that is prone to change.


nathangoulding:
This is probably not necessary since `status_code` is exported but if you want it feel free to re-open.

Issue #288, opened 2015-10-27
Oauth2.0_token() does not work servers requiring client credentials in the header.


tkinsf:
httr's oauth2 works fine with google or gibhub. But it does not work with fitbit. Fitbit requires the client crdentials in the "Authorization" header and oauth2.0_token() seems to pass them as parameters rather than in the header. According to oauth2.0 rfc section 2.3.1, supporting the authorization header is a MUST for the server while supporting them as parameters is a MAY. As such, it would be better/safer for client to pass the client id/secret in the header when requesting the access token.

hadley:
Please provide a reproducible example. `oauth2.0_token()` defaults to `as_header = TRUE`, so I'm pretty sure they're already sent in the header by default.

tkinsf:
Hadly,

The documentation says that as_header determines weather the oauth is
passed in bearer header or as a parameter.  So, I assumed it meant how the
token is passed, not the client credentials in authorization header while
obtaining the token itself. Am I wrong?

I don't have definitive evidence that oauth2.0_token passes client
id/secret as parameters. (I was gonna packet-trace, then I realized that
packets were encoded..). What I know is that fitbit service works fine with
my google script that explicitly passes id/secret in the authorization
header while httr gets "authorization header required" response.

Let me know if you know a way to definitively prove. I'll be happy to
experiment, but my R knowledge is too limited to delve into the library
source code itself..

Thanks
tk

hadley:
If you don't want to look at the source, start by providing a reproducible example.

tkinsf:
Here you go, Hadly. Attached file contains the code, and you can log in to
my fitbit account as "sung.tk@gmail.com"/"Convex425". Let me know if you
need anything else.

tk

tkinsf:
Oops, it is "sung.tk@gmail.com"/"Sanse425". Sorry.

grahamrp:
I also have a problem accessing the fitbit api. I've looked at the https requests in [Fiddler](http://www.telerik.com/fiddler) and can confirm that the request to obtain the access token by `oauth2.0_token` does not include the client credentials in the header. Looking at the source, the `as_header = TRUE` argument to `oauth2.0_token` does not affect the way the request for the token is made, it only affects subsequent calls to the api *once the token has been obtained*.

The request for the token is made by `init_oauth2.0` via a `POST` *without HTTP Basic authentication* and with the `client_secret` in the body:

```R
# Use authorisation code to get (temporary) access token
[...]

PR #310, opened 2016-01-01
Add option to retrieve oauth2.0 token using basic authentication

grahamrp:
Regarding issue #288 this adds an option to `oauth2.0_token` so that when the authorization code is exchanged for the token it can be done using http basic authentication, rather than the default method of including the app key and secret in the request body.

From the [OAuth2.0 Framework RFC](https://tools.ietf.org/html/rfc6749) it looks like authorization servers should provide the http basic authentication method as a minimum, and the method of providing the app key/secret in the request body is optional. I see that the github api allows both, but fitbit only allows basic authentication.

### DIFF
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -51,11 +51,16 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
 #'     code. Defaults to the of the \code{"httr_oob_default"} default,
 #'     or \code{TRUE} if \code{httpuv} is not installed.
 #' @param is_interactive Is the current environment interactive?
+#' @param use_basic_auth if \code{TRUE} use http basic authentication to
+#'     retrieve the token. Some authorization servers require this.
+#'     If \code{FALSE}, the default, retrieve the token by including the
+#'     app key and secret in the request body.
 #' @export
 #' @keywords internal
 init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
                           use_oob = getOption("httr_oob_default"),
-                          is_interactive = interactive()) {
+                          is_interactive = interactive(),
+                          use_basic_auth = NULL) {
   if (!use_oob && !is_installed("httpuv")) {
     message("httpuv not installed, defaulting to out-of-band authentication")
     use_oob <- TRUE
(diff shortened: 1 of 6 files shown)


R/oauth-init.R
```diff
@@ -51,11 +51,16 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
...
+#'     retrieve the token. Some authorization servers require this.
+#'     If \code{FALSE}, the default, retrieve the token by including the
+#'     app key and secret in the request body.
 #' @export
 #' @keywords internal
 init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
                           use_oob = getOption("httr_oob_default"),
-                          is_interactive = interactive()) {
+                          is_interactive = interactive(),
+                          use_basic_auth = NULL) {
```
hadley: It would be more clear to make the default value `FALSE`

R/oauth-init.R
```diff
@@ -85,13 +90,23 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
...
 
   # Use authorisation code to get (temporary) access token
-  req <- POST(endpoint$access, encode = "form",
-    body = list(
-      client_id = app$key,
-      client_secret = app$secret,
-      redirect_uri = redirect_uri,
-      grant_type = "authorization_code",
-      code = code))
+  if (isTRUE(use_basic_auth)){
```
hadley: Could you add a comment with link to the relevant section of the RFC?

R/oauth-token.r
```diff
@@ -181,8 +181,9 @@ Token1.0 <- R6::R6Class("Token1.0", inherit = Token, list(
 #' caching policies used to store credentials across sessions.
 #'
 #' @inheritParams init_oauth2.0
-#' @param as_header If \code{TRUE}, the default, sends oauth in bearer header.
-#'   If \code{FALSE}, adds as parameter to url.
+#' @param as_header If \code{TRUE}, the default, configures the token to add
+#' itself to the bearer header of subsequent requests.
```
hadley: Missing indent

hadley:
Looks good, apart from a few minor comments.  Could you please also a bullet to NEWS?

R/oauth-init.R
```diff
@@ -85,13 +90,25 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
...
-  req <- POST(endpoint$access, encode = "form",
-    body = list(
-      client_id = app$key,
-      client_secret = app$secret,
-      redirect_uri = redirect_uri,
-      grant_type = "authorization_code",
-      code = code))
+  # Send credentials using HTTP Basic or as parameters in the request body
+  # See https://tools.ietf.org/html/rfc6749#section-2.3 (Client Authentication)
+  if (isTRUE(use_basic_auth)){
```
hadley: It feels like there's a little too much duplication here. Could you please pull out the common `body`?

grahamrp:
I've pulled out the common body now, thanks. I've also moved the NEWS bullet to the top. I hope the commits are okay - I'm a novice at collaboration on github.

hadley:
Thanks!

PR #312, opened 2016-01-05
Add option user_params to oauth2.0_token to enable endpoint-specific access parameters.

cornf4ke:
This branch holds the implementation for passing additional access parameters in the form of a list of named values to the oauth2.0_token function. The function refresh_oauth2.0 was modified accordingly.

The service account functionality was not touched. My only reference for the use of user_params, Microsoft Azure AD, uses quite different parameters for service-to-service authorization when compared to httr's init_oauth_service_account. See https://msdn.microsoft.com/en-us/library/azure/dn645543.aspx.

A demo for the Azure AD endpoint was added. In the committed version it uses dummy credentials and will run but not succeed.

While viewing the diffs on github, I noticed a possible typo in the demo/00Index line
"oauth2-linkedin     Using linkedin api with OAuth 1.0" which probably should read "OAuth 2.0".

### DIFF
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -46,6 +46,8 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
 #' @inheritParams init_oauth1.0
 #' @param type content type used to override incorrect server response
 #' @param scope a character vector of scopes to request.
+#' @param user_params List of named values holding endpoint specific parameters to pass to
+#'     the server when posting the request for obtaining or refreshing the access token.
 #' @param use_oob if FALSE, use a local webserver for the OAuth dance.
 #'     Otherwise, provide a URL to the user and prompt for a validation
 #'     code. Defaults to the of the \code{"httr_oob_default"} default,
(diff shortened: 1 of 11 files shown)


R/oauth-init.R
```diff
@@ -85,13 +87,16 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
...
 
   # Use authorisation code to get (temporary) access token
-  req <- POST(endpoint$access, encode = "form",
-    body = list(
-      client_id = app$key,
-      client_secret = app$secret,
-      redirect_uri = redirect_uri,
-      grant_type = "authorization_code",
-      code = code))
+  req_params <- c(
```
hadley: I think it would be slightly better to use `stats::modifyList()` and change the order of the argument so that `user_params` "loses" if a name is defined in both places.

R/oauth-init.R
```diff
@@ -85,13 +87,16 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL, type = NULL,
...
+      list(
+          client_id = app$key,
+          client_secret = app$secret,
+          redirect_uri = redirect_uri,
+          grant_type = "authorization_code",
+          code = code
+      ),
+      user_params)
+  req <- POST(endpoint$access, encode = "form", body=req_params)
```
hadley: Need a space around last `=`

hadley:
Apart from a couple of minor things, looks good.  Could you please also add a bullet point to NEWS?

cornf4ke:
Thanks for your comments. I added the news bullet and modified the code according to your notes.

PR #313, opened 2016-01-05
fix for token service account

nathangoulding:
@hadley This fixes the root cause of the `Error: is.request(y) is not TRUE` that you've been seeing.

It was only breaking when using the service token, and is caused by the `sign` method not returning what `request_combine` is expecting. PTAL.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,6 +36,8 @@
 
 * `build_url()` now collapses vector `path` with `/` (#280, @artemklevtsov).
 
+* Ensure that OAuth token `sign()` returns a `request` (#313, @nathangoulding).
+
 # httr 1.0.0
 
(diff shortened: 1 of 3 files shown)


hadley:
This looks good! Two requests:

* Could you please add a bullet point to news?
* Could you add an explicit `stopifnot(is.request(x))` in the right place to make sure we don't accidentally introduce this problem in the future? (i.e. if we add more token classes)

nathangoulding:
Great, I think I've addressed these.

hadley:
Thanks!

For future reference, it's a bit better to put news items at the top - it seems to generate fewer spurious merge conflicts.

nathangoulding:
Ahh OK, roger that.

PR #315, opened 2016-01-06
clear token on oauth2.0 error

### LINKED ISSUE #311: Error after manually revoking OAuth permissions
This has occurred when using bigrquery, but I believe this to be an httr issue.

I revoked bigrquery permission for my Google account from https://myaccount.google.com/security#connectedapps. After this, each attempt to execute a query gives the following message (even after restarting R):

```
> query_exec(sql, project = project)
Auto-refreshing stale OAuth token.
Error in refresh_oauth2.0(self$endpoint, self$app, self$credentials) : 
  client error: (400) Bad Request
9: stop(http_condition(x, "error", call = sys.call(-1)))
8: stop_for_status(response)
7: refresh_oauth2.0(self$endpoint, self$app, self$credentials)
6: req$auth_token$refresh()
5: request_perform(req, hu$handle$handle)
4: POST(paste0(base_url, url), body = json, config(token = token), 
       add_headers(`Content-Type` = "application/json"), ..., query = prepare_bq_query(query)) at request.r#31
3: bq_post(url, body) at jobs.r#60
2: insert_query_job(query, project, destination_table = destination_table, 
       default_dataset = default_dataset) at query.r#32
1: query_exec(sql, project = project)
```

nathangoulding:
This implements OAuth2.0 error handling [as described in the RFC](https://tools.ietf.org/html/rfc6749#section-5.2) on refresh, and clears the token locally.

Fixes #311. BigQuery [respects the OAuth2.0 RFC](https://cloud.google.com/bigquery/troubleshooting-errors), so the behavior he describes is resolved without any changes to `bigrquery`.

Tested locally both using the local file cache and not.


krlmlr:
This works better. On the first try, I'm seeing:

```
> query_exec(sql, project = project)
Auto-refreshing stale OAuth token.
No encoding supplied: defaulting to UTF-8.
Error: Invalid Credentials
In addition: Warning message:
In req$auth_token$refresh() :
  Unable to get refreshed credentials with refresh token
9: stop(out$err$message, call. = FALSE) at request.r#67
8: process_request(req) at request.r#34
7: bq_post(url, body) at jobs.r#60
6: insert_query_job(query, project, destination_table = destination_table, 
       default_dataset = default_dataset) at query.r#32
5: query_exec(sql, project = project) at oauth.R#4
4: eval(expr, envir, enclos)
3: eval(ei, envir)
2: withVisible(eval(ei, envir))
1: source("oauth.R")
```

Only the second try opens a browser for re-authentication. Furthermore, from RStudio, I was able to connect even after I had revoked the access from the Google console and after restarting RStudio. Then I tried from the R console, which led to the behavior above. I'm not sure if this is due to a timeout or some magic cache.

R/oauth-error.r
```diff
@@ -0,0 +1,25 @@
...
+  "unsupported_grant_type",
+  "invalid_scope"
+)
+
+# This implements error checking according to the OAuth2.0
+# specification: https://tools.ietf.org/html/rfc6749#section-5.2
+has_oauth2.0_error <- function(response) {
+  if (status_code(response) %in% oauth2.0_error_codes) {
+    content <- content(response)
+    if (content["error"] %in% oauth2.0_errors) {
```
hadley: That should be `content$error` or `content[["error"]]`

hadley: But it seems like it would be safer to assume that any error implies that it failed.

R/oauth-refresh.R
```diff
@@ -17,6 +17,9 @@ refresh_oauth2.0 <- function(endpoint, app, credentials) {
   )
 
   response <- POST(refresh_url, body = body, encode = "form")
+  if (has_oauth2.0_error(response)) {
+    return(NULL)
```
hadley: I would rather this threw an error, to be consistent with the `stop_for_status()` below

hadley: But maybe you're trying to separate this into known errors and unknown errors? I'd like to hear your thinking about this.

nathangoulding: My thinking was that here in `refresh_oauth2.0` our goal is to get new, valid credentials using our refresh token (obviously), and OAuth2 provides a defined mechanism to inform us of that this can't happen - by returning `invalid_grant` in the error response.

Separating this out from `stop_for_status` allows us to still catch any errors that occur - 404s, redirects, internal server errors, etc. All of these are errors that prevent us from getting refreshed credentials, but there isn't any specific information in them that tells us what to do.

Contrast that with the OAuth2 error(s) that tell us we have an invalid token, which is why our logic is to clear the token from our cache.

Taking it a step further, the OAuth2 spec is pretty clear that `invalid_grant` is the error we should look for in the case of the refresh token. I've currently written this in such a way that `has_oauth2.0_error` will return `TRUE` even in the case of OAuth2 errors that aren't specific to the refresh token.

I did opt to cast a broader net, but there's an argument to be made that this function should be called `has_oauth2.0_refresh_token_error(response)` and only return `TRUE` when `invalid_grant` is present.

Thoughts?

R/oauth-token.r
```diff
@@ -211,7 +211,12 @@ Token2.0 <- R6::R6Class("Token2.0", inherit = Token, list(
   refresh = function() {
     self$credentials <- refresh_oauth2.0(self$endpoint, self$app,
         self$credentials, self$params$user_params)
-    self$cache()
```
hadley: I think the logic here could be made more clear. Maybe:

```R
creds <- refresh_oauth2.0(self$endpoint, self$app,
  self$credentials, self$params$user_params)
if (is.null(creds) {
  remove_cached_token(self)
  warning("Unable refresh token", call. = FALSE)
} else {
  self$credentials <- creds
  self$cache()
}
self
```

?

R/oauth-error.r
```diff
@@ -0,0 +1,25 @@
...
+  "invalid_client",
+  "invalid_grant",
+  "unauthorized_client",
+  "unsupported_grant_type",
+  "invalid_scope"
+)
+
+# This implements error checking according to the OAuth2.0
+# specification: https://tools.ietf.org/html/rfc6749#section-5.2
+has_oauth2.0_error <- function(response) {
```
hadley: Maybe call this `known_oauth2_error()`?

R/oauth-token.r
hadley: Or maybe the warning should occur in `refresh_oauth2.0()`?

nathangoulding: Yeah, I think the warning should probably occur in `refresh_oauth2.0` in thinking about it.

nathangoulding:
PTAL

krlmlr:
I'm now reliably seeing the following behavior (using nathangoulding/bigrquery#71):

- Initial state: Connection from bigrquery works, no service token set
- Revoke access via https://myaccount.google.com/security#connectedapps
    - I can still connect from bigrquery for a minute or so
- Eventually, I'm seeing the following as a result from query_exec():
    ```
    Auto-refreshing stale OAuth token.
    No encoding supplied: defaulting to UTF-8.
    Error: Invalid Credentials
    In addition: Warning message:
    Unable to refresh token 
    9: stop(out$err$message, call. = FALSE) at request.r#67
    8: process_request(req) at request.r#34
    7: bq_post(url, body) at jobs.r#60
    6: insert_query_job(query, project, destination_table = destination_table, 
           default_dataset = default_dataset) at query.r#32
    5: query_exec(sql, project = project) at oauth.R#5
    4: eval(expr, envir, enclos)
    3: eval(ei, envir)
    2: withVisible(eval(ei, envir))
    1: source("~/git/R/bigrquery/oauth.R", echo = TRUE)
    ```
- The second call of query_exec() finally opens the authentication page in the browser

hadley:
@nathangoulding LTGM - can you please add a bullet to NEWS?

hadley:
@krlmlr I think eliminating the second refresh will require a tweak in bigrquery

nathangoulding:
Updated NEWS.

hadley:
Thanks!

PR #316, opened 2016-01-08
oauth1.0 rsa-sha1 signature support

### LINKED ISSUE #307: OAuth RSA-SHA1 signatures
[JIRA](https://developer.atlassian.com/jiradev/jira-apis/jira-rest-apis/jira-rest-api-tutorials/jira-rest-api-example-oauth-authentication) incoming OAuth1.0 authentication only works with RSA-SHA1 signatures and httr only has HMAC-SHA1 at the moment, making integration difficult. It would be nice if httr knew how to make RSA-SHA1 signatures.

A specialised Token$sign() might be possible but it's far from pretty.

### LINKED ISSUE #308: HTTP POST for OAuth1.0 token requests
[RFC5849](http://tools.ietf.org/html/rfc5849#section-2.1) suggests that token requests should be obtained using POST unless the server demands something else. [Core 1.0a](http://oauth.net/core/1.0a/#rfc.section.6.3.1) also recommended POST. httr currently uses GET, which complicates integration with servers that only work with POST (e.g. JIRA).

It would be nice if the endpoint or app configuration could select HTTP methods for obtaining request tokens (and perhaps access tokens).

nathangoulding:
Implements RSA-SHA1 signature support for OAuth1.0. Assuming you've created an application link [according to this documentation](https://developer.atlassian.com/jiradev/jira-apis/jira-rest-apis/jira-rest-api-tutorials/jira-rest-api-example-oauth-authentication) with your local RSA private key or `key.pem`:

```r
endpoint <- oauth_endpoint(
  request = "https://yourjira.atlassian.net/plugins/servlet/oauth/request-token",
  authorize = "https://yourjira.atlassian.net/plugins/servlet/oauth/authorize",
  access = "https://yourjira.atlassian.net/plugins/servlet/oauth/access-token")
app <- oauth_app("jira", key = "yourkey", secret = "yoursecret")
key <- openssl::read_key("/path/to/key.pem")
token <- oauth1.0_token(endpoint, app, private_key = key)
GET("https://yourjira.atlassian.net/rest/api/2/search?jql=assignee=fred", config(token = token))
```

Enables support for authenticating against JIRA. Fixes #307. Fixes #308.


nathangoulding:
Switched to use `openssl` after looking back at the commits and realizing `PKI` wasn't being used at all anymore.

R/oauth-init.R
```diff
@@ -10,15 +10,16 @@
...
 
-  oauth_sig <- function(url, method, token = NULL, token_secret = NULL, ...) {
-    oauth_header(oauth_signature(url, method, app, token, token_secret, ...,
+  oauth_sig <- function(url, method, token = NULL, token_secret = NULL, private_key = NULL, ...) {
+    oauth_header(oauth_signature(url, method, app, token, token_secret, private_key, ...,
       callback = oauth_callback()))
   }
 
   # 1. Get an unauthorized request token
-  response <- GET(endpoint$request, oauth_sig(endpoint$request, "GET"))
```
hadley: I'm slightly concerned whether this change is safe in general, but I tried it out with four oauth1 demos, and it seemed to be ok

R/rsa.r
```diff
@@ -0,0 +1,16 @@
+#' RSA SHA1
+#'
+#' As described in \url{http://tools.ietf.org/html/rfc3447}.
+#'
+#' @param key secret key
+#' @param string data to securely hash
+#' @keywords internal
+#' @export
+rsa_sha1 <- function(key, string) {
```
hadley: I think this should go in the same file (and possibly be documented with) `hmac_sha1`

R/oauth-init.R
nathangoulding: Yeah, I'm hyper sensitive to not wanting to introduce regressions with this change, though the [RFC5849](http://tools.ietf.org/html/rfc5849#section-2.1) and [its predecessor](http://oauth.net/core/1.0a/#rfc.section.6.3.1) are fairly prescriptive that it should be `POST` unless otherwise noted:

```
   The client obtains a set of temporary credentials from the server by
   making an authenticated (Section 3) HTTP "POST" request to the
   Temporary Credential Request endpoint (unless the server advertises
   another HTTP request method for the client to use).
```

And:

```
To request an Access Token, the Consumer makes an HTTP request to the
Service Provider's Access Token URL. The Service Provider documentation
specifies the HTTP method for this request, and HTTP POST is RECOMMENDED.
The request MUST be signed per Signing Requests, and contains the
following parameters:
```

R/rsa.r
nathangoulding: Sounds good. Right now `hmac_sha1` is in its own file `hmac.r`. Should I move both to `oauth-signature.r`, or move both to a new common `sha1.r` file?

R/oauth-init.R
nathangoulding: Is it worth it to introduce `request_method = "POST"` and `access_method = "POST"` to `oauth_endpoint()` potentially?

R/rsa.r
hadley: I'd say common `sha1.r`

R/oauth-init.R
hadley: Given that the RFC prescriptive and the change works for four implementations in the wild, I'd say it's fine to change the default.  Let's not worry about making it optional until someone actually complains about it.

hadley: I'm going to start the release process either later today or on Tuesday. I'll give a two week notice, and explicitly encourage people to test it if they're using OAuth.

R/oauth-init.R
```diff
@@ -110,7 +111,7 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL, user_params = NULL,
     req <- POST(endpoint$access, encode = "form", body = req_params,
       authenticate(app$key, app$secret, type = "basic"))
   } else {
-    body$client_secret <- app$secret
+    req_params$client_secret <- app$secret
```
nathangoulding: It seems like there was a regression introduced with the basic auth PR. If you'd rather this be in its own PR just let me know.

nathangoulding:
I think I've addressed your comments and roxygenned.

R/oauth-init.R
hadley: That was probably my bad merge conflict handling :/

hadley:
LGTM - can you please add a bullet to NEWS?

DESCRIPTION
```diff
@@ -27,4 +27,4 @@ Suggests:
 VignetteBuilder: knitr
 License: MIT + file LICENSE
 URL: https://github.com/hadley/httr
-RoxygenNote: 5.0.1
+RoxygenNote: 5.0.1.9000
```
hadley: I try to avoid using dev roxygen with other packages, but it's not a big deal.

nathangoulding: Ahh OK, my bad. I have to resolve the conflict with NEWS so I'll revert this change as well.

nathangoulding:
OK all set with NEWS and DESCRIPTION. Let me know if you want me to close this, rebase, and re-open a new PR to clean up the commit log.

hadley:
Thanks!

hadley:
Ooops, just noticed you forgot to document `private_key`. Could you please do a new PR?

nathangoulding:
Ahh crap, yeah one sec.

PR #333, opened 2016-02-08
close connection when done reading

jeroenooms:
Bug found by @jennybc. Currently `httr` leaves open lingering connections after uploading a file:

```r
library(httr)
citation <- upload_file(system.file("CITATION"))
req <- POST("http://httpbin.org/post", body = citation)
showConnections()
```

So you get the annoying warning:

```r
rm(req)
gc()
# Warning message:
# closing unused connection 3 (/Library/Frameworks/R.framework/Versions/3.2/Resources/library/base/CITATION) 
```

It looks like there was a subtle change in behavior of libcurl. Old versions keep calling `readfunctions` until it returns 0 bytes. But the recent libcurl assumes we are done once readfunction returns less than the requested data.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr 1.1.0.9000
 
+* Fix in readfunction to close connection when done.
+
 # httr 1.1.0
 
 ## New features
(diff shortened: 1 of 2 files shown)


jennybc:
With the new testthat, each of these unused connections now also causes a warning during testing, e.g.

```
...
list sheets: .W.....
permissions: ....W...........
register sheets: ...........
upload sheets: ..WWWW...
...
```

which is going to be very sad on, e.g., travis.

PR #337, opened 2016-02-12
Change encoding for query part; fixes #335

### LINKED ISSUE #335: encoding/decoding '+' as space (rather than literal '+' sign)
Without fully understanding RFC 1808, I think that in the query section of the URI that the '+' should not be translated to '%2B', which is a literal '+' sign, but to '%20', which is a space character.  For instance, you can see this:

    > library(httr)
    > 
    > url <- "http://google.com?q=saint+paul+mn"
    > 
    > parse_url(url)
    $scheme
    [1] "http"

    $hostname
    [1] "google.com"

    $port
    NULL

    $path
    [1] ""

    $query
    $query$q
    [1] "saint+paul+mn"

    $params
    NULL

    $fragment
    NULL

    $username
    NULL

    $password
    NULL

    attr(,"class")
    [1] "url"
    > 
    > build_url(parse_url(url))
    [1] "http://google.com/?q=saint%2Bpaul%2Bmn"

If you compare the original `url` to the parsed/built url, they generate different queries.  In the first case, the '+' signs are interpreted as spaces, but after parse/build, a literal '+' sign is passed.  The current behavior breaks some use cases.

I do not know if this is actually in compliance with RFC 1808 and there needs to be some pre-processing that needs to happen to make this not break, or it is not in compliance due to the meaning of the '+' in the query portion of the URI.

Also, I'll reference this in terms of this discussion: http://stackoverflow.com/a/29948396/1022967

jennybc:
Not sure this is the best solution. But it confirms the query needs to be encoded differently than other parts of the URL. It rescues my example from #335.

``` r
library(httr)

query_space <- "type:issue author:hadley repo:rstudio/rstudioapi state:open"
query_plus <- "type:issue+author:hadley+repo:rstudio/rstudioapi+state:open"
list_space <- list(q = query_space)
list_plus <- list(q = query_plus)

(url <-
  modify_url("https://api.github.com", path = "search/issues", query = list_space))
#> [1] "https://api.github.com/search/issues?q=type:issue%20author:hadley%20repo:rstudio/rstudioapi%20state:open"
status_code(x <- GET(url))
#> [1] 200

(url <-
  modify_url("https://api.github.com", path = "search/issues", query = list_plus))
#> [1] "https://api.github.com/search/issues?q=type:issue+author:hadley+repo:rstudio/rstudioapi+state:open"
status_code(x <- GET(url))
#> [1] 200
```

### DIFF
--- a/R/body.R
+++ b/R/body.R
@@ -40,7 +40,8 @@ body_config <- function(body = NULL, encode = "form", type = NULL)  {
 
   # Deal with three ways to encode: form, multipart & json
   if (encode == "form") {
-    body_raw(compose_query(body), "application/x-www-form-urlencoded")
+    body_raw(compose_query(body, part = "other"),
+             "application/x-www-form-urlencoded")
   } else if (encode == "json") {
     body_raw(jsonlite::toJSON(body, auto_unbox = TRUE), "application/json")
   } else if (encode == "multipart") {
--- a/R/url-query.r
+++ b/R/url-query.r
@@ -7,23 +7,27 @@ parse_query <- function(query) {
   values
 }
 
-compose_query <- function(elements) {
+compose_query <- function(elements, part = c("query", "other")) {
   if (length(elements) == 0)
     return("")
 
   if (!all(has_name(elements)))
     stop("All components of query must be named", call. = FALSE)
 
+  part <- match.arg(part)
+
   stopifnot(is.list(elements))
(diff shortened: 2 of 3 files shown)


hadley:
Again, I'm sceptical that this is correct. I'm happy to be proven wrong, but you'll need to point to the relevant RFCs

PR #349, opened 2016-03-31
Expose na parameter of jsonlite::toJSON in PATCH, POST, PUT and VERB

kosmoz:
This helps the package to cover some more use cases. Since the default value stays the same, there should not arise any issues.

Now the parameter `json.NA` may be specified when calling `PATCH`, `POST`, `PUT` or `VERB`. The parameter must be `"null"` or `"string"`. The default is `c("null", "string")`. Example:
```
result = httr::POST(url, body=content, encode = "json", 
  					json.NA = "null")
```


hadley:
It doesn't feel right to add an argument that only affects json output. Can you please file as an issue and I'll think about an approach that feels better to me.

Issue #356, opened 2016-04-10
httr should not use "customrequest" (CURLOPT_CUSTOMREQUEST)


antoine-lizee:
This has a lot of undesired effect, including the inability of `httr` and `rvest` to properly follow redirects.

By setting [here](https://github.com/hadley/httr/blob/7261a525ede1beb9880d1ce2e2009d27c8e8bef7/R/request.R#L102) `CURLOPT_CUSTOMREQUEST` to `"POST"` or any other method instead of using `CURLOPT_POST`, `CURLOPT_PUT` ... and friends, the behavior of all the subsequent queries following the request will be modified, including the redirects. 

The bug that caught me (and that I saw on SO in several obscure flavors) is to login successfully following a POST, and then crash (get a 403 or something related) following the ensued redirect, because the redirect issues a `POST` again instead of the ubiquitous `GET`. (also, see https://github.com/hadley/httr/issues/65)

FYI, even though the RFC actually mentions that on `301`, `302` and `303`, the redirect method [should not change](https://tools.ietf.org/html/rfc2616#section-10.3.3), this is the default behaviour of all browsers, and of cURL as well - unless you specify the `CURLOPT_POSTREDIR` [option](https://curl.haxx.se/libcurl/c/CURLOPT_POSTREDIR.html), but this is another story.

antoine-lizee:
Small remark: because the `customrequest` option is last to be set at the request level, it's basically impossible to revert this behavior:
``` 
httr::POST(........, httr::config(customrequest = "", post = TRUE))
```
...will have no effect.

PR #357, opened 2016-04-10
enh: use the appropriate CURL options for specifying methods

antoine-lizee:
cf #356.
A small fix can go a long way :-). I'd advocate for quick publishing to CRAN after inclusion of this PR (or whatever other code change that solves #356).

### DIFF
--- a/R/request.R
+++ b/R/request.R
@@ -99,7 +99,14 @@ print.request <- function(x, ...) {
 
 request_prepare <- function(req) {
   req <- request_combine(request_default(), req)
-  req$options$customrequest <- req$method
+  switch(req$method,
+         "POST" = {
+           req$options$post <- TRUE},
+         "PUT" = {
+           req$options$put <- TRUE},
+         "GET" = {
+           req$options$httpget <- TRUE},
+         {req$options$customrequest <- req$method})
 
   # Sign request, if needed
   token <- req$auth_token
--- a/R/request.R
+++ b/R/request.R
@@ -99,13 +99,13 @@ print.request <- function(x, ...) {
 
 request_prepare <- function(req) {
   req <- request_combine(request_default(), req)
+
+  # Use the appropriate cURL method when available.
+  # This - among others - ensures proper behaviour on redirects.
   switch(req$method,
-         "POST" = {
-           req$options$post <- TRUE},
-         "PUT" = {
-           req$options$put <- TRUE},
-         "GET" = {
-           req$options$httpget <- TRUE},
+         "POST" = {req$options$post <- TRUE},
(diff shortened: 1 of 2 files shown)


R/request.R
```diff
@@ -99,7 +99,14 @@ print.request <- function(x, ...) {
 
 request_prepare <- function(req) {
   req <- request_combine(request_default(), req)
-  req$options$customrequest <- req$method
+  switch(req$method,
+         "POST" = {
```
hadley: Could you please make it more like `"POST" = (req$options$post <- TRUE)` ?

hadley:
Would you mind also adding a bullet point to NEWS?

And any thoughts on how to test this?

R/request.R
antoine-lizee: Why the brackets instead of curly? I'd rather have it be silent?

antoine-lizee:
NEWS -> Will do
test -> Well, the problem came up with a wrong redirect behavior (see #356), and as you pointed out quite some time ago (cf https://github.com/Runscope/httpbin/issues/156), this is un-testable with httpbin (POSTing to `/redirect` throws a `405 METHOD NOT ALLOWED`).

R/request.R
hadley: I'm not sure what you mean by silent, but I don't think choice of parens matters.

hadley:
Since testing is hard, could you add a comment mentioning why this behaviour is important? I'm worried I'll forget about it and re-simplify in the future.

R/request.R
antoine-lizee: Curly brackets are expressions, they return the result invisibly. Just try:
```
switch("a", "a" = (tt <- 2))
# [1] 2
switch("a", "a" = {tt <- 2})
# Nothing
```

hadley: Yes, but it really doesn't matter here. Regardless, splitting the expression over multiple lines really bugs me, so please fix.

hadley:
Thanks!

hadley:
Unfortunately this doesn't seem to be correct - setting `PUT` (which has been deprecated since curl 7.12.0) causes PUT requests to hang for forever on windows. Setting `UPLOAD`, which is the recommend option, cause it to hang forever on my computer too.

I think this is probably because I'm setting the post body parameters in `body_config()`.

hadley:
A bit of googling suggests that using `CUSTOMREQUEST` is the correct way to do PUTs in libcurl

antoine-lizee:
Good catch - this is officially a Windows exception to the cURL behaviour, itself being a standard but non-compliant way of dealing with RFC... Why is software getting so crazy? 
Thanks!

PR #364, opened 2016-05-11
Allow to easily customize number of digits for JSON serialization

kirillseva:
by setting `options(httr.digits = desired_number_of_digits)`


kirillseva:
No idea why travis is failing, trivial change and tests pass locally

hadley:
I think it is a really really really bad idea to use options in this way.

kirillseva:
@hadley What do you think would be a good interface for exposing this flag from jsonlite to httr? I'm willing to make it right, seems like a useful feature to have

hadley:
It doesn't really seem that useful to me.  I'd rather add a new encoding type, say "raw", that let you encode the json yourself so you could do whatever you want with it.

PR #371, opened 2016-05-24
Set cache permissions to 0600

### LINKED ISSUE #365: `.httr-oauth` file should be created with perms `0600` by default
Title says it all -- I can't see a reason that we shouldn't default to creating the file with restrictive permissions. I'm happy to send a PR if folks agree.

hadley:
Fixes #365

@craigcitro can you PTAL?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # httr 1.1.0.9000
 
+* The cross-session OAuth cache is now created with permission 0600, and should
+  give a better error if it can't be created (#365).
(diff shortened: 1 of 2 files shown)


craigcitro:
LGTM

Only remaining question: do we want to try chmod'ing existing cache files that we run into? (I don't think it's worth it, just wanted to say it out loud.)

hadley:
I think that'd also get tricky if someone had deliberately set the value to something else.

PR #372, opened 2016-05-24
Implement RETRY()

### LINKED ISSUE #353: Need function for retrying a request
* Fixed number of tries
* Exponential back off
* Built-in randomness

hadley:
This keeps retrying until a request succeeds (in a server friendly way)

Fixes #353 

@craigcitro can you PTAL?


R/retry.R
```diff
@@ -0,0 +1,58 @@
...
+#' below 400). It is designed to be kind to the server: after each failure
+#' randomly waits up to twice as long. (Technically it uses exponential
+#' backoff with jitter, using the approach outlined in
+#' \url{https://www.awsarchitectureblog.com/2015/03/backoff.html.})
+#'
+#' @inheritParams VERB
+#' @inheritParams GET
+#' @inheritParams POST
+#' @param times Maximum number of requests to attempt.
+#' @param pause_base,pause_cap This method uses exponential back-off of with
```
craigcitro: two nits: 
 * spurious `of`
 * space after `pause_base`? (not actually sure if you want one)

R/retry.R
```diff
@@ -0,0 +1,58 @@
...
+#'   to use \code{\link{stop_for_status}()}.
+#' @export
+#' @examples
+#' # Succeeds straight away
+#' RETRY("GET", "http://httpbin.org/status/200")
+#' # Never succeeds
+#' RETRY("GET", "http://httpbin.org/status/404")
+RETRY <- function(verb, url = NULL, config = list(), ...,
+                  body = NULL, encode = c("multipart", "form", "json", "raw"),
+                  times = 3, pause_base = 1, pause_cap = 60,
```
craigcitro: You probably want to validate `pause_base` and `pause_cap` here, so that they're checked before the first failing request.

R/retry.R
```diff
@@ -0,0 +1,58 @@
...
+                  handle = NULL, quiet = FALSE) {
+  stopifnot(is.numeric(times), length(times) == 1L)
+
+  hu <- handle_url(handle, url, ...)
+  req <- request_build(verb, hu$url, body_config(body, match.arg(encode)), config, ...)
+  resp <- request_perform(req, hu$handle$handle)
```
craigcitro: slightly silly, but you push this line into the for loop below, a la

    i <- 1
    while (i <= times && http_error(resp)) {
      ...
    }

This has two advantages:
* if you do end up wanting any more validation on the response, you only have to do it once
* you'll actually sleep between the first two requests, which you don't currently. 😉

craigcitro:
LGTM, just a few comments above.

Worth adding a little test?

R/retry.R
hadley: Roxygen adds the space automatically in the generated docs

R/retry.R
hadley: But don't you need an additional check? `i <= times && (is.null(resp) || http_error(resp))` ?   (I think you actually want a do-while loop here, but R doesn't have that)

hadley:
Comments incorporated. Any thoughts on how to test?

PR #373, opened 2016-06-01
Do not prepend oauth_ to additional parameters for oauth_signature

jimhester:
The flickr API for instance needs all query parameters to be included in
the signature. (https://www.flickr.com/services/api/auth.oauth.html#yui_3_11_0_1_1464787470505_317)

With this change I am able to get a 200 response with the following code (app secret redacted)

```r
f <- oauth_app("r to flickr", "ed724990662ac2b8c236726a34226ea9", "<redacted>")

ep <- oauth_endpoint(
  request = "https://www.flickr.com/services/oauth/request_token",
  authorize = "https://www.flickr.com/services/oauth/authorize",
  access = "https://www.flickr.com/services/oauth/access_token"
)

tok <- oauth1.0_token(ep, f, cache = T)

signature <- oauth_signature("https://api.flickr.com/services/rest", app =
  tok$app, token = tok$credentials$oauth_token, token_secret =
  tok$credentials$oauth_token_secret,
  other_params = list(method = "flickr.test.login", format = "json", "nojsoncallback" = 1))

GET("https://api.flickr.com/services/rest", query = signature)
```

### DIFF
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -16,8 +16,8 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
                           private_key = NULL) {
 
   oauth_sig <- function(url, method, token = NULL, token_secret = NULL, private_key = NULL, ...) {
-    oauth_header(oauth_signature(url, method, app, token, token_secret, private_key, ...,
-      callback = oauth_callback()))
+    oauth_header(oauth_signature(url, method, app, token, token_secret, private_key,
+        other_params = c(list(...), oauth_callback = oauth_callback())))
   }
 
   # 1. Get an unauthorized request token
@@ -31,11 +31,11 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
   authorize_url <- modify_url(endpoint$authorize, query = list(
     oauth_token = token,
     permission = "read"))
-  verifier <- oauth_listener(authorize_url, is_interactive)[[1]]
+  verifier <- oauth_listener(authorize_url, is_interactive)$oauth_verifier
 
   # 3. Request access token
   response <- POST(endpoint$access,
-    oauth_sig(endpoint$access, "POST", token, secret, verifier = verifier, private_key = private_key),
+    oauth_sig(endpoint$access, "POST", token, secret, oauth_verifier = verifier, private_key = private_key),
     body = ""
   )
   stop_for_status(response)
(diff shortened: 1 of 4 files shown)


R/oauth-init.R
```diff
@@ -31,11 +31,11 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
   authorize_url <- modify_url(endpoint$authorize, query = list(
     oauth_token = token,
     permission = "read"))
-  verifier <- oauth_listener(authorize_url, is_interactive)[[1]]
+  verifier <- oauth_listener(authorize_url, is_interactive)$oauth_verifier
```
hadley: I think the use of position was deliberate here

R/oauth-init.R
```diff
@@ -31,11 +31,11 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
...
 
   # 3. Request access token
   response <- POST(endpoint$access,
-    oauth_sig(endpoint$access, "POST", token, secret, verifier = verifier, private_key = private_key),
+    oauth_sig(endpoint$access, "POST", token, secret, oauth_verifier = verifier, private_key = private_key),
```
hadley: Are you sure it doesn't need to be `oauth_private_key`?

R/oauth-init.R
jimhester: Yeah it was changed with https://github.com/hadley/httr/commit/7261a525ede1beb9880d1ce2e2009d27c8e8bef7#commitcomment-17702567, unfortunatly subsetting by position breaks when authenticating with flickr at least. I don't know if they are not returning the proper response or if it needs to be reverted.

R/oauth-init.R
jimhester: private_key is a named parameter to the `oauth_sig()` function defined above and doesn't get passed in the new other_params argument. It looks like it is used at https://github.com/hadley/httr/pull/373/files#diff-196777e55557c984260eab22d25ac411L45 for the "RSA-SHA1" method rather than as an oauth parameter.

hadley: Oh ooops, yeah.

R/oauth-init.R
jimhester: Ok I have updated it as suggested with https://github.com/hadley/httr/pull/373/commits/da2bef023f93f44e0c6166359fe88d0369cd7aad

jimhester:
Ok this was modified as discussed to check for oauth_verifier in the returned JSON and otherwise take the first item.

PR #383, opened 2016-07-13
RETRY min pause

r2evans:
**Use case**: retry connections where sub-second delays are acceptable, perhaps only advisable with local connections. Will print a warning unless the user is doubly-sure they want to use sub-second delays by adding `pause_min_warning=FALSE`.


hadley:
Why do you need this?

r2evans:
I have a polling client that pushes/pulls data, and failures are frequent-enough that I needed to solidify the recovery process while minimizing the likelihood of giving up. The software was doing relatively high-frequency (100s of milliseconds) work, so having a 1sec delay is debilitating. I recognize that sub-second retries can be traumatic to the server (ergo a sane default of 1 second minimum), but I know what is causing the transient failures and am comfortable enough going shorter on the retry interval.

hadley:
Ok, that makes sense. Can you please merge/rebase and then remove `pause_min_warning` and the warning? That feels like overkill to me.  

This function has become sufficiently complicated that it probably needs a complete overhaul and a more flexible API. That'll need to wait until httr2.

r2evans:
Done.

hadley:
Yeah, something has gone horribly wrong 😢

r2evans:
Oh crap, I see it now ... sigh, I really need to do better with merge/rebase, apparently I understand it even less than I thought I did (which says a lot). I'll kill this and put up a new PR with a new diff, sorry 'bout that.

r2evans:
@hadley off topic (and I'm borderline "questionable" using this venue for this question), do you have a personal preference on reference/tutorial/guidance for dealing with `rebase`-vs-`merge` in collaborating-`git` workflows? I thought I was following good-practice with this branch and would much prefer to not have to go through this rigamarole every time.

hadley:
I don't sorry 😦

jennybc:
Sorry to butt in uninvited, but @r2evans it might help to know that repos owned by r-lib and tidyverse tend to use "squash and merge" on pull requests. So the overall diff is absolute top priority, not the journey by which you got there. That might eliminate one source of angst re: how to sync up a PR with master.

r2evans:
Thanks @jennybc, it helps (and I encourage "butting-in" on this question). Bottom line for me is that I may end up forensically tracing a bad `rebase` so I can see when it went wrong. Some tutorials touch on this but don't go into the weeds. For now I will focus on being more careful, and likely only using `merge` on PRs I've already pushed up.

jennybc:
This seems like a decent resource: http://willi.am/blog/2014/08/12/the-dark-side-of-the-force-push/

I would probably merge for a PR that I'd already pushed and had generated a bit of discussion. If I did rebase, I would also force push. But it is much easier to do bad things with that workflow.

r2evans:
Good reads, thanks. As those articles suggest, I am loath to "force push": I don't want to seriously mess up somebody else's repo. Yet I am learning.

PR #385, opened 2016-07-20
two small copy edits

hansthompson:
Hello Hadley.  Thanks for making this package.  

I would like to make the switch to httr from XML but for my package in development, [rusps](https://github.com/hansthompson/rusps).  Please give it a try if you like.  I think address validation is an important data science tool and this package is for using the free USPS address validating api.  I would love you criticism/thoughts/pull requests.


hadley:
Nice - thanks!

Issue #384, opened 2016-07-14
Oauth2 Client Credentials Grant not possible


cderv:
Hi, 

I tried using `httr` with Oauth2  authorization on some API of my company. This API is working with the _Client Credentials Grant_ as described [here](https://tools.ietf.org/html/rfc6749#section-4.4). It is one of the 4 grant type for Oauth2.

In this grant type, no authorization request is needed, only access token requests are made. 
Access request are made with `client id` and `client secret`, in an authorization header. Nothing is passed as request parameters (as there is no authorization code).  Request should looks like this : 

```
`POST /token HTTP/1.1
     Host: server.example.com
     Authorization: Basic czZCaGRSa3F0MzpnWDFmQmF0M2JW
     Content-Type: application/x-www-form-urlencoded

     grant_type=client_credentials
```

First, I tried to use `oauth2.0_token` but it did not work. It took me a while to understand why there was a problem. It seems that `oauth2.0_token` works for a certain type of grant where authorization is needed. `init_oauth2.0` calls in its workflow `oauth_exchanger` or `oauth_listener` on a authorize url. This step of authentification request is not optional, which is not compatible with  the [client credendial grant](https://tools.ietf.org/html/rfc6749#section-4.4).

The other issue is related to the fix of #288 with client credential in the header. Option `use_basic_auth` works but there is still a body passed in the request:

``` r
if (isTRUE(use_basic_auth)) {
      req <- POST(
        endpoint$access,
        encode = "form",
        body = req_params,
        authenticate(app$key, app$secret, type = "basic")
      )
    }
```

`req_params` contains some parameters not needed in the _client credential grant_, in particular their is no `code` to pass. 

Knowing all that, in order to use `httr` Oauth dance, I extended the Token2.0 ref class to implement a new `initialize` method that calls a modified `init_oauth2.0` function in which I just commented the `authorize` request part, and call POST with no `code` param in the body. It's a quick fix I did today but it works. 

So, 
- Did I missed something and it should have worked in the first place ? 
- Is it a good idea to create a new extended class for that ? 
- Should it be rather a fix in the `init_oauth2.0` function as an option to get an access token without any authorization request and code beforehand ? like `use_basic_auth` arg in #288 which offer a choice in the method.

I just experiment it today but I'll be glad to share my works and findings if it is any interest.

jennybc:
[...]

PR #388, opened 2016-07-31
Client credential grant made possible

cderv:
As explained in #384 , Oauth2 credential grant just need an access token request and then is not possible with `httr` oauth2 mechanism because authorization request is made at every call. 

In this PR, I just add an option to skip the part with an authorization request in order to make use of API with Oauth2 credential grant possible. 

Optional arg `without_auth_req` is set to `FALSE` by default to keep everyting else the same and not break anything.

---

This change is [Reviewable](https://reviewable.io/reviews/hadley/httr/388)

### DIFF
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -62,12 +62,19 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
 #'     retrieve the token. Some authorization servers require this.
 #'     If \code{FALSE}, the default, retrieve the token by including the
 #'     app key and secret in the request body.
+#' @param without_auth_req if \code{TRUE}, no authorization request is send
+#'   before token access request. It is suitable for the \emph{Client Credential
+#'   Grant} as described in
+#'   \url{https://tools.ietf.org/html/rfc6749#section-4.4}. If \code{FALSE}, the
+#'   default, authorization request is then token request is send included the
+#'   authorization code.
 #' @export
 #' @keywords internal
 init_oauth2.0 <- function(endpoint, app, scope = NULL, user_params = NULL,
                           type = NULL, use_oob = getOption("httr_oob_default"),
                           is_interactive = interactive(),
-                          use_basic_auth = FALSE) {
+                          use_basic_auth = FALSE,
+                          without_auth_req = FALSE) {
   if (!use_oob && !is_installed("httpuv")) {
     message("httpuv not installed, defaulting to out-of-band authentication")
     use_oob <- TRUE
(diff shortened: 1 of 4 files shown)


R/oauth-init.R
```diff
@@ -105,18 +105,26 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL, user_params = NULL,
...
-    code = code)
 
-  if(without_auth_req) req_params <- NULL
+
+  if(without_auth_req) {
+    req_params <- list(
+      client_id = app$key,
+      client_secret = app$secret,
+      # redirect_uri = redirect_uri,
+      grant_type = "client_credentials")
```
cderv: As tested in another use case with Yelp API (see #384), we may not have to use this `if` solution to set `grant_type` as `grant_type = "client_credentials"` could be pass manually by users through `user_params = list(grant_type = "client_credentials")`. 
However, there is still `code = code` that we do not need a a *client credentials grant* type

hadley:
Rather than to continue overloading `oauth2.0_token` with more arguments, it might be better to create a new function. Which of the existing arguments to `oauth2.0_token` also apply to client credential grants?

cderv:
Client credential grant do not have an authorize phase. So no authorize url to visit for a obtaining a code. It this part of `init_oauth2.0` that need to be optional for this kind of grant. In the issue discussion, @jennybc  gave an example where she needed to put a dummy authorize url and blank response to skip this part and get the token. It is why I added a new arg to skip this part like `use_basic_auth` or `use_oob`.

I can get a token using 

```r
rte_app <- oauth_app("opendatarte", key = Sys.getenv("RTE_ID"),
                      secret = Sys.getenv("RTE_SECRET"))
```
I need to put a dummy authorize for this to work but It does not require one, like for the yelp example
```r
rte_endpoint <-
  oauth_endpoint(NULL,
                 authorize = "https://digital.iservices.rte-france.com/token/oauth/",
                 access = "https://digital.iservices.rte-france.com/token/oauth/")
```
then I need `use_oob = T` to have the message in the console and not open a browser window. I need to hit enter interactively for the *oauth dance* to continue and obtaining the token. This API uses basic auth so I need `use_basic_auth = T` and I need to precise the grant type.
```r
token <- oauth2.0_token(rte_endpoint, rte_app,
                        user_params = list(grant_type = "client_credentials"),
                        use_basic_auth = T, use_oob = T)
```
This work to obtain the token.

I do not know for the Yelp API but in my case I do not have a refresh endpoint for the token. 

so this give an error
```r
token$refresh()
# Error: Refresh token not available
```

I needed to define a new R6 Class for this to work asking for a new token using `init_oauth2.0` for credential grant. May be another issue but related to this grant.
```r
TokenDataRTE <- R6::R6Class("TokenDataRTE", inherit = httr:::Token2.0, list(
  can_refresh = function() {
    TRUE
  },
  refresh = function() {
    cred <- httr::init_oauth2.0(
      endpoint = self$endpoint,
      app = self$app,
      user_params = self$params$user_params,
      use_basic_auth = self$params$use_basic_auth,
      without_auth_req = self$params$without_auth_req
    )
    if (is.null(cred)) {
      httr:::remove_cached_token(self)
    } else {
      self$credentials <- cred
      self$cache()
      .state$token$credentials <- cred
    }
    self
  }))
```
Does it get you more insight on this case ?
How should I modify this PR ? or write another one ?

hadley:
I think the main question is whether we want to do a major refactoring now, or wait until httr2.  It's probably better to wait, and reconsider the API from the ground-up after a careful reading of RFC 6749. 

For now, lets make the argument `client_credentials`, defaulting to `FALSE`. Can you please also tweak the documentation for `oauth_endpoint` saying to set `authorize` to `NULL` if not needed.

R/oauth-init.R
hadley: I think this block could be made simpler by setting `code <- NULL` if `client_credentials`. 

Then you could write:

```R
req_params <- list(
  client_id = app$key,
  redirect_uri = redirect_uri,
  grant_type = if (client_credentials) "client_credentials" else "authorization_code",
  code = code
)
```

cderv:
OK. Thanks. I see where you want me to go. 
Should I rebase the PR first ? or make a new one? What are your advices on an old PR with conflicts ?

R/oauth-init.R
cderv: For client credential, there is no `redirect_uri` too. I need to put something like that for this to work :
```r
  req_params <- list(
    client_id = app$key,
    redirect_uri = if (client_credentials) NULL else redirect_uri,
    grant_type = if (client_credentials) "client_credentials" else "authorization_code",
    code = code
  )
```
Code is set to `NULL` by previous code.

hadley:
Unless you're pretty familiar with git, it's probably just easier to start with a fresh branch and PR.

cderv:
I did a rebase with the current branch, I did handle merge conflict but I need to force push. not sure how a PR discussion respond to that. 

So, I began a new branch. It is in my repo.  will make a PR when it is ready. Still need to tweak the documentation and add bullet NEWS. Will do this when I will come back Sunday. 

Thanks for your help.

hadley:
Sounds great - thanks!  I'll close this PR and look forward to your new one.

PR #406, opened 2016-11-03
Auto-close browser page after successful OAuth

krlmlr:
Only if JavaScript is enabled.

http://stackoverflow.com/a/12791753/946850

### DIFF
--- a/R/oauth-listener.r
+++ b/R/oauth-listener.r
@@ -42,8 +42,8 @@ oauth_listener <- function(request_url, is_interactive = interactive()) {
 
     list(
       status = 200L,
-      headers = list("Content-Type" = "text/plain"),
-      body = "Authentication complete. Please close this page and return to R."
+      headers = list("Content-Type" = "text/html"),
+      body = readChar(system.file("oauth/close.html", package = utils::packageName()), 1e6)
     )
   }
   use <- listener_endpoint()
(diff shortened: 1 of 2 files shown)


saurfang:
Yes please. One less friction. Any concerns?

krlmlr:
Works well for me in Chromium on Ubuntu. Don't know about other browsers/OSs.

jennybc:
It doesn't seem to be working for me (Mac OS / Safari 10.1) 😞. This *would* be really nice.

hadley:
Won't this just close the tab, leaving the user in the browser with no idea what to do next?

krlmlr:
That's true. I guess we could add a close button and an "auto-close all future such windows" checkbox.

@jennybc: All this PR does is showing an HTML file with a tiny bit of JavaScript. Can you post/show a self-closing HTML document that works on your system?

jennybc:
@krlmlr If I visit an html file with this snippet, it does indeed lead to the associate Safari tab being closed after a short delay.

```
<script type="text/javascript">setTimeout("window.close();", 3000);</script>
```

so I am not sure why my trial of this PR does not lead to autoclose of the tab where I completed OAuth. I just checked one more time, though, and it still doesn't close the tab.

krlmlr:
Strange. When testing this PR, did you notice an alert by Safari that JavaScript is blocked? What happens if you start Safari from the command line and pass it the HTML file from this PR?

yutannihilation:
This doesn't work for my Firefox. While it seems possible to close windows in Chrome, Firefox is still too secure to close window by script:
https://stackoverflow.com/a/19768082/5397672

hadley:
This seems too fragile for minimal gain. I think the current behaviour needs to stay.

PR #409, opened 2016-11-13
Implement request callbacks, closes #408

### LINKED ISSUE #408: Better API client testing
## Motivation

One way to test API clients without performing thousands of requests all the time, is to record the requests and responses and then replay them while testing. There are tools for this in other languages as well. I wrote a very simple tool for R and `httr`: https://github.com/gaborcsardi/httrmock

Unfortunately, without modifications in `httr` itself, `httrmock` is hacky and fragile, it relies on tracing `httr` functions. With some minimal support from `httr` we could make it much better, and CRAN-grade proper.

## The callback option

The idea is that `httr` would get a callback option, and if this is turned on, then the callback function can determine whether to perform the actual HTTP requests. 

The callback function would be called (if installed) both at the beginning of `request_perform` and at the end, and `httr` would pass both the HTTP `request` and the `response` to the callback. `httrmock and possibly other packages can then implement API client testing, and HTTP request recording and replaying easily. 

Sg like this would go to the beginning of `request_perform`:

```r
if (! is.null(httr_callback)) {
  cbres <- httr_callback(mode = "request", req)
  if (!is.null(cbres)) return(cbres)
}
```

The callback is called with the request, and if it returns non-`NULL`, then we terminate early, without actually performing the request. This is *replay* mode, and the callback must return a proper response in this case.

If the callback returns `NULL`, then the request is performed.

Sg like this would go to the end of `request_perform`:

```r
res <- response(...)  etc.
if (! is.null(httr_callback)) {
  callback(mode = "response", req, res)
}
res
```

This is to allow the *recording* of the request.

Please let me know if this is something you would consider adding to `httr` and then I can put together a simple PR. Thanks!

gaborcsardi:
More details in #408. There is an HTTP mocking implementation that uses this, in the https://github.com/gaborcsardi/httrmock package.


R/request.R
```diff
@@ -122,6 +122,12 @@ request_prepare <- function(req) {
 
 request_perform <- function(req, handle, refresh = TRUE) {
   stopifnot(is.request(req), inherits(handle, "curl_handle"))
+
+  orig_req <- req
```
hadley: Can you this and the block below up into a small helper function that lives in `callback.r`?

gaborcsardi: Sure. Btw. I was also thinking that probably we should use the request as it is sent out, i.e. after `request_prepare`, no?

gaborcsardi:
Yes, two callbacks is definitely cleaner, I'll do that, too.

R/request.R
hadley: Yes, I'd say so.

gaborcsardi:
@hadley OK, updated, PTAL.

hadley:
LGTM - can you please add a bullet to NEWS?

gaborcsardi:
Added NEWS.

hadley:
Thanks!

Issue #363, opened 2016-04-25
Add reddit example


hadley:
```R
library(httr)
library(purrr)

app <- oauth_app("reddit", "bvmjj2EOBvOknQ", "n8ueSvTNdlE0BDDJpLljvmgUGUw")
reddit <- oauth_endpoint(
  authorize = "https://www.reddit.com/api/v1/authorize",
  access = "https://www.reddit.com/api/v1/access_token"
)

token <- oauth2.0_token(reddit, app, scope = "read modposts", use_basic_auth = TRUE)
```

muschellij2:
This is currently giving
```r
Error in init_oauth2.0(self$endpoint, self$app, scope = self$params$scope,  : 
  Too Many Requests (RFC 6585) (HTTP 429). Failed to get an access token.
```

I think it's because the `User-Agent` is using the default from Reddit, which gets throttled quickly.  Is it possible to add `User-Agent` to the header of `oauth2.0_token`?  I can't pass using `user_params`, even if `as_header = TRUE`.

PR #411, opened 2016-11-21
added config argument for oauth2.0 and init_oauth

muschellij2:
This refers to https://github.com/hadley/httr/issues/363 where the reddit oauth2.0 example does not work unless you can add to the header.  In the case of reddit, you need to add the user agent, and the resulting example/demo works:

```r
app <- oauth_app("reddit", "bvmjj2EOBvOknQ", "n8ueSvTNdlE0BDDJpLljvmgUGUw")
reddit <- oauth_endpoint(
  authorize = "https://www.reddit.com/api/v1/authorize",
  access = "https://www.reddit.com/api/v1/access_token"
)
token <- oauth2.0_token(reddit, app, 
  scope = "read modposts", 
  use_basic_auth = TRUE,
  config = user_agent("redditr_agent"))
```

whereas the example does not (429 too many requests, because default user-agent is over-used):
```r
app <- oauth_app("reddit", "bvmjj2EOBvOknQ", "n8ueSvTNdlE0BDDJpLljvmgUGUw")
reddit <- oauth_endpoint(
  authorize = "https://www.reddit.com/api/v1/authorize",
  access = "https://www.reddit.com/api/v1/access_token"
)
token <- oauth2.0_token(reddit, app, 
  scope = "read modposts", 
  use_basic_auth = TRUE)
```


R/oauth-token.r
```diff
@@ -197,10 +197,12 @@ oauth2.0_token <- function(endpoint, app, scope = NULL, user_params = NULL,
                            type = NULL, use_oob = getOption("httr_oob_default"),
                            as_header = TRUE,
                            use_basic_auth = FALSE,
-                           cache = getOption("httr_oauth_cache")) {
+                           cache = getOption("httr_oauth_cache"),
+                           config = list()) {
```
hadley: I think this needs a more specific name because there are other places in the OAuth dance that could have config params. Maybe `config_init`?

hadley:
Fixes #413

muschellij2:
Should be sufficient to merge after checks are done.  

* Fix reddit demo where `user_agent` was being overloaded.  Now can pass `user_agent` using `config_init` to `init_oauth2.0` and `Token2.0$new()`  (@muschellij2 @hadley #363 and #413).

* Changed example in `write_stream` from https://jeroenooms.github.io/data/diamonds.json to https://github.com/jeroen/data/raw/gh-pages/diamonds.json as this had moved (@muschellij2).

Also, fixed an example on `write_stream`: https://github.com/r-lib/httr/blob/1078b1d35d0c5d0a61d53461cc6c37f9fb49551d/R/write-function.R#L61 as that link was dead and causing errors in the "examples" part of R CMD check

hadley:
Did you close this deliberately?

muschellij2:
Just meant to comment, not close and comment.

John

hadley:
Maybe it's better to resubmit anyway - your PR now has quite a few spurious changes and two concerns. That makes it hard to review efficiently.

muschellij2:
Agreed - it was a few months ago so I think I may have some gaps there.
I'll check and resubmit

John

PR #433, opened 2017-03-23
Switch round order

jeroen:
Need this for a future version of `curl`. Simply switches around `if` options, no side effects.


hadley:
Thanks!

PR #435, opened 2017-04-18
OAuth 1.0 normalized URL must exclude default port

leeper:
There is a bug in `oauth_signature()` related to port handling. The [OAuth 1.0 spec](https://oauth.net/core/1.0/#anchor14) indicates:

> Unless specified, URL scheme and authority MUST be lowercase and include the port number; http default port 80 and https default port 443 MUST be excluded.

In other words, if port is specified as 80 for HTTP or 443 or HTTPS, the port number needs to be excluded from the normalized URL that is used in the OAuth 1.0 signature. `oauth_signature()` currently always includes the port number even if it is the default port number.

This PR fixes that.

### DIFF
--- a/R/oauth-signature.r
+++ b/R/oauth-signature.r
@@ -49,6 +49,13 @@ oauth_signature <- function(url, method = "GET", app, token = NULL,
   method <- toupper(method)
 
   url <- parse_url(url)
+  if (!is.null(url$port)) {
+    if (url$scheme == "http" && url$port == "80") {
+      url$port <- NULL
+    } else if (url$scheme == "https" && url$port == "443") {
+      url$port <- NULL
+    }
+  }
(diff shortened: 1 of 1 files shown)


hadley:
Would you mind also adding a test? I think the easiest way would be to split `oauth_signature` into two pieces, one that generates `oauth` and the other that encodes it.

hadley:
@leeper are you interested in finishing this off?

leeper:
Sorry, I wanted to get to it but haven't had time. Maybe this weekend?

hadley:
I'd like to release by Friday, but I can finish it off.

leeper:
Ok, thanks. Sorry for not getting to it.

hadley:
No problems. Refactoring was a bit more involved than I'd expected so probably better for me to do anything. Thanks for bringing this to my attention and for the fix!

PR #437, opened 2017-04-27
Expand git and Rbuild ignoring of newly created oauth cache files; fixes #436

### LINKED ISSUE #436: New oauth cache files are not added to .gitignore, .Rbuildignore
If `cache = TRUE` or `cache = "some/path"` (vs `NA`) in `httr::oauth2.0_token()`, a newly created `.httr-oauth` file is NOT in fact added to `.gitignore` and `.Rbuildignore`. This is contrary to the documentation and, I think?, behaviour at some point in the past.

The failure to `.gitignore` is the bigger problem and  has lead to some googlesheets users committing and pushing tokens to GitHub (googlesheets sets `cache = TRUE` by default).

jennybc:
Fixes #436

httr is only adding a new cache file to `.gitignore` and `.Rbuildignore` in the case of `cache = NA`. This PR makes it do so when `cache = TRUE` and when `cache` holds a path.

I am doing each test re: oauth caching in a fresh temp directory because behaviour is affected by which files already exist and it's unsettling to have code like `unlink(".gitignore")`. I also save and restore options in these tests, because `use_cache()` potentially changes the value of `httr_oauth_cache`.

### DIFF
--- a/tests/testthat/test-oauth-cache.R
+++ b/tests/testthat/test-oauth-cache.R
@@ -57,15 +57,14 @@ test_that("token saved to and restored from cache", {
 test_that("new caches are Rbuildignored and gitignored", {
   owd <- setwd(tmp_dir())
   on.exit(setwd(owd))
-  old <- options()
-  on.exit(options(old), add = TRUE)
   file.create("DESCRIPTION")
 
   ## default: options("httr_oauth_cache" = NA)
   ## not tested
   ## if cache does not exist, will not be created if !interactive()
 
-  options("httr_oauth_cache" = TRUE)
+  old <- options("httr_oauth_cache" = TRUE)
+  on.exit(options(old), add = TRUE)
   use_cache()
   expect_true(file.exists(".Rbuildignore"))
   expect_identical(readLines(".Rbuildignore"), "^\\.httr-oauth$")
--- a/tests/testthat/test-oauth-cache.R
+++ b/tests/testthat/test-oauth-cache.R
@@ -1,6 +1,17 @@
 context("OAuth cache")
 
+tmp_dir <- function() {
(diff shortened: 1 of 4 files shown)


tests/testthat/test-oauth-cache.R
```diff
@@ -39,3 +53,30 @@ test_that("token saved to and restored from cache", {
   expect_equal(token_b$endpoint, token_a$endpoint)
   expect_equal(token_b$credentials, token_a$credentials)
 })
+
+test_that("new caches are Rbuildignored and gitignored", {
+  owd <- setwd(tmp_dir())
+  on.exit(setwd(owd))
+  old <- options()
```
hadley: Why not do `old <- options("httr_oauth_cache" = TRUE)` ?

jennybc: I just copied the sole pre-existing instance of this pattern verbatim. I can do this  ☝️ instead.

jennybc: done 70c29fdbba1245cfbe1821c537848383441f0db5

hadley:
Thanks - can you please add a bullet to NEWS?

hadley:
Thanks!

jennybc:
I added the bullet. Also explicitly checked whether this was a regression. I don't think it was. I think oauth cache have been ignore only when `httr_oauth_cache = NA` and user approves creation of `.httr-oauth` interactively.

I still think this PR is needed. It seems correct and makes the docs true:

> The cache file should not be included in source code control or R packages (because it contains private information), so httr will automatically add the appropriate entries to `.gitignore` and `.Rbuildignore` if needed.

PR #439, opened 2017-05-04
Fix typos in vignettes

bbrewington:
Remember to press F7 when you're done editing a Vignette in RStudio ✔️ :)

Hopefully none of my corrections are forcing US English...if that's the case, feel free to ignore.


hadley:
Thanks!

PR #448, opened 2017-06-01
close txtProgressBar after reaching total

cjyetman:
Pretty sure this is what was originally intended... looks like some curly braces got left out somehow. `close(bar)` seems adequate for wrapping things up (closing adds a newline), but the `utils::flush.console()` may have had other/additional purpose?

### DIFF
--- a/R/progress.R
+++ b/R/progress.R
@@ -46,8 +46,10 @@ progress_bar <- function(type, con) {
         first <<- FALSE
       }
       cat("\rDownloading: ", bytes(now, digits = 2), "     ", sep = "", file = con)
-      if (now == total) cat("\n", file = con)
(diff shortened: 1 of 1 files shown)


R/progress.R
```diff
@@ -46,8 +46,10 @@ progress_bar <- function(type, con) {
         first <<- FALSE
       }
       cat("\rDownloading: ", bytes(now, digits = 2), "     ", sep = "", file = con)
-      if (now == total) cat("\n", file = con)
```
hadley: This code needs to remain as is - this is for the case where curl doesn't know the total download size. 

I think you should add this code below `setTxtProgressBar` so two branches of the if are parallel.

PR #450, opened 2017-06-16
Add mime types text/delim and text/csv2

paulponcet:


hadley:
These don't appear to be official mime types? They're not listed in https://www.iana.org/assignments/media-types/media-types.xhtml

PR #452, opened 2017-06-22
Don't form empty queries; closes #451

### LINKED ISSUE #451: Empty but non-NULL query
Distilling a conversation with @craigcitro from another channel.

This is about empty queries, i.e. where the query list is not `NULL` but it also has no content. Should there be a trailing `?`? I think that's up for debate, but in my "for discussion" PR (#452), I said NO. I'm more certain that there should never be a trailing `?=`.

Examples of what httr currently does with some empty queries:

``` r
library(httr)
modify_url("http://google.com", query = list(a = NULL))
#> [1] "http://google.com/?="
modify_url("http://google.com", query = list())
#> [1] "http://google.com/?"
```

Section 3.3 of <https://www.ietf.org/rfc/rfc1738.txt> seems to be the most relevant and it doesn't seem definitive to me.

```
The HTTP URL scheme is used to designate Internet resources
   accessible using HTTP (HyperText Transfer Protocol).

   The HTTP protocol is specified elsewhere. This specification only
   describes the syntax of HTTP URLs.

   An HTTP URL takes the form:

      http://<host>:<port>/<path>?<searchpart>

   where <host> and <port> are as described in Section 3.1. If :<port>
   is omitted, the port defaults to 80.  No user name or password is
   allowed.  <path> is an HTTP selector, and <searchpart> is a query
   string. The <path> is optional, as is the <searchpart> and its
   preceding "?". If neither <path> nor <searchpart> is present, the "/"
   may also be omitted.
```

For context, @craigcitro offers this link re: how Go handles empty queries (TL;DR it's configurable via an option): https://golang.org/pkg/net/url/#URL

jennybc:
See #451 

Before:

``` r
library(httr)
modify_url("http://google.com", query = list(a = NULL))
#> [1] "http://google.com/?="
modify_url("http://google.com", query = list())
#> [1] "http://google.com/?"
```

After:

``` r
devtools::load_all()
#> Loading httr
modify_url("http://google.com", query = list(a = NULL))
#> [1] "http://google.com/"
modify_url("http://google.com", query = list())
#> [1] "http://google.com/"
```

### DIFF
--- a/R/url.r
+++ b/R/url.r
@@ -161,5 +161,3 @@ modify_url <- function(url, scheme = NULL, hostname = NULL, port = NULL,
 
   build_url(utils::modifyList(old, new))
 }
-
-compact <- function(x) Filter(Negate(is.null), x)
--- a/R/url.r
+++ b/R/url.r
@@ -120,11 +120,11 @@ build_url <- function(url) {
   }
 
   if (is.list(url$query)) {
-    query <- compose_query(url$query)
+    query <- compose_query(compact(url$query))
   } else {
     query <- url$query
   }
-  if (!is.null(query)) {
+  if (!is.null(query) && nzchar(query)) {
     stopifnot(is.character(query), length(query) == 1)
     query <- paste0("?", query)
   }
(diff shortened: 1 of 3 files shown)


hadley:
Ooops, I merged slightly too soon. Would you mind doing another PR to add a bullet to news?

PR #453, opened 2017-06-30
Fix overly sensitive unit test (error messages might fluctuate)

jeroen:
The new version of `curl` has some more detailed error messages with slightly different wording.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ r:
   - release
   - devel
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,6 +9,9 @@ r:
   - release
   - devel
 
+r_github_packages:
```
hadley: Can you use remotes instead please?

PR #458, opened 2017-07-24
Add bullet re: recognizing more empty queries

jennybc:
as requested here https://github.com/r-lib/httr/pull/452#issuecomment-317343268


hadley:
Thanks!

PR #459, opened 2017-07-25
Improvements to RETRY (#404)

asieira:
* Add new `terminate_on` parameter that allows status_codes that
prevent retries to be specified;
* Catch error conditions using `tryCatch` and retry if they occur.

### DIFF
--- a/R/retry.R
+++ b/R/retry.R
@@ -81,11 +81,13 @@ backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60, quiet =
   length <- ceiling(stats::runif(1, max = min(pause_cap, pause_base * (2 ^ i))))
   if (!quiet) {
     if (inherits(resp, "error")) {
-      error_description <- gsub("[\n\r]+$", "", as.character(resp))
-      message(error_description, "\nRequest failed [ERROR]. Retrying in ", length, " seconds...")
+      error_description <- gsub("[\n\r]*$", "\n", as.character(resp))
+      status <- "ERROR"
     } else {
-      message("Request failed [", status_code(resp), "]. Retrying in ", length, " seconds...")
+      error_description <- ""
+      status <- status_code(resp)
     }
+    message(error_description, "Request failed [", status, "]. Retrying in ", length, " seconds...")
   }
   Sys.sleep(length)
 }
--- a/R/retry.R
+++ b/R/retry.R
@@ -80,8 +80,12 @@ retry_should_terminate <- function(i, times, resp, terminate_on) {
 backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60, quiet = FALSE) {
   length <- ceiling(stats::runif(1, max = min(pause_cap, pause_base * (2 ^ i))))
   if (!quiet) {
-    status <- if (inherits(resp, "error")) as.character(resp) else status_code(resp)
-    message("Request failed [", status, "]. Retrying in ", length, " seconds...")
+    if (inherits(resp, "error")) {
+      error_description <- gsub("[\n\r]+$", "", as.character(resp))
+      message(error_description, "\nRequest failed [ERROR]. Retrying in ", length, " seconds...")
+    } else {
+      message("Request failed [", status_code(resp), "]. Retrying in ", length, " seconds...")
+    }
   }
   Sys.sleep(length)
 }
--- a/R/retry.R
+++ b/R/retry.R
@@ -21,7 +21,7 @@
 #' @param quiet If \code{FALSE}, will print a message displaying how long
 #'   until the next request.
 #' @param terminate_on Optional vector of numeric HTTP status codes that if found
-#'   on the response will immediately stop the retries. If missing or NULL, will
+#'   on the response will terminate the retry process. If missing or NULL, will
 #'   keep retrying while \code{\link{http_error}()} is \code{TRUE} for the response.
 #' @return The last response. Note that if the request doesn't succeed after
 #'   \code{times} times this will be a failed request, i.e. you still need
--- a/R/retry.R
+++ b/R/retry.R
@@ -1,7 +1,11 @@
 #' Retry a request until it succeeds.
 #'
-#' Safely retry a request until it succeeds (returns an HTTP status code
-#' below 400). It is designed to be kind to the server: after each failure
+#' Safely retry a request until it succeeds, as defined by the \code{terminate_on}
+#' parameter, which by default means a response for which \code{\link{http_error}()}
+#' is \code{FALSE}. Will also retry on error conditions raised by the underlying curl code,
+#' but if the last retry still raises one, \code{RETRY} will raise it again with
+#' \code{\link{stop}()} to maintain backwards compatibility.
+#' It is designed to be kind to the server: after each failure
 #' randomly waits up to twice as long. (Technically it uses exponential
 #' backoff with jitter, using the approach outlined in
 #' \url{https://www.awsarchitectureblog.com/2015/03/backoff.html}.)
@@ -16,6 +20,10 @@
 #'   \code{pause_cap} seconds.
 #' @param quiet If \code{FALSE}, will print a message displaying how long
 #'   until the next request.
+#' @param terminate_on Optional vector of numeric or integer HTTP status codes
+#'   that if found on the response will immediately stop the retries. If missing
+#'   or NULL, will keep retrying while \code{\link{http_error}()} is \code{TRUE}
+#'   for the response.
 #' @return The last response. Note that if the request doesn't succeed after
 #'   \code{times} times this will be a failed request, i.e. you still need
 #'   to use \code{\link{stop_for_status}()}.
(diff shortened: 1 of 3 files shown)


R/retry.R
```diff
@@ -16,6 +20,10 @@
 #'   \code{pause_cap} seconds.
 #' @param quiet If \code{FALSE}, will print a message displaying how long
 #'   until the next request.
+#' @param terminate_on Optional vector of numeric or integer HTTP status codes
```
hadley: Integer is a sub-type of numeric, so you don't need to mention both here (or test for both below)

R/retry.R
```diff
@@ -16,6 +20,10 @@
...
+#'   that if found on the response will immediately stop the retries. If missing
```
hadley: ... that will terminate the retry process.

It's redundant to mention the missing behaviour since there's a default argument.

R/retry.R
```diff
@@ -25,32 +33,55 @@
...
+                  handle = NULL, quiet = FALSE, terminate_on = NULL) {
   stopifnot(is.numeric(times), length(times) == 1L)
   stopifnot(is.numeric(pause_base), length(pause_base) == 1L)
   stopifnot(is.numeric(pause_cap), length(pause_cap) == 1L)
+  stopifnot(is.integer(terminate_on) || is.numeric(terminate_on) || is.null(terminate_on))
 
   hu <- handle_url(handle, url, ...)
   req <- request_build(verb, hu$url, body_config(body, match.arg(encode)), config, ...)
-  resp <- request_perform(req, hu$handle$handle)
+  resp <- tryCatch(request_perform(req, hu$handle$handle), error = function(e) e)
```
hadley: I don't love having a variable called `resp` that sometimes contains a response and sometimes contains an error object. However, I can't see an obviously better API.

R/retry.R
```diff
@@ -25,32 +33,55 @@
...
+  while (!retry_should_stop(i, times, resp, terminate_on)) {
+    backoff_full_jitter(i, resp, pause_base, pause_cap, quiet = quiet)
 
     i <- i + 1
-    resp <- request_perform(req, hu$handle$handle)
+    resp <- tryCatch(request_perform(req, hu$handle$handle), error = function(e) e)
   }
 
-  resp
+  if ("error" %in% class(resp)) {
```
hadley: Should use `inherits()` (and below)

R/retry.R
```diff
@@ -25,32 +33,55 @@
...
+    stop(resp)
+  } else {
```
hadley: I think you can drop the else block, and simply return `resp` (but without the explicit return)

R/retry.R
```diff
@@ -25,32 +33,55 @@
...
+    return(resp)
+  }
+}
+
+retry_should_stop <- function(i, times, resp, terminate_on) {
+  if (i >= times) {
+    return(TRUE)
```
hadley: Please remove the explicit returns

R/retry.R
```diff
@@ -25,32 +33,55 @@
...
+  } else {
+    return(!http_error(resp))
+  }
 }
 
-backoff_full_jitter <- function(i, status, pause_base = 1, pause_cap = 60, quiet = FALSE) {
+backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60, quiet = FALSE) {
   length <- ceiling(stats::runif(1, max = min(pause_cap, pause_base * (2 ^ i))))
   if (!quiet) {
+    status <- if ("error" %in% class(resp)) as.character(resp) else status_code(resp)
```
hadley: I don't think coercing the error to a string here is the right approach because it's going to be completely different to a short http status code. Instead it would be be better to just insert "ERROR" or similar.

R/retry.R
```diff
@@ -25,32 +33,55 @@
...
 
-  resp
+  if ("error" %in% class(resp)) {
+    stop(resp)
+  } else {
+    return(resp)
+  }
+}
+
+retry_should_stop <- function(i, times, resp, terminate_on) {
```
hadley: I think should be `retry_should_terminate` to be consistent with the argument name

R/retry.R
asieira: I think the default value of NULL is not clear on what status codes will be terminated on the default scenario. I thought about setting the default value of `terminate_on` to `400:599`, but the problem is that this is not exactly the same as running `http_error` (which will check for `>= 400`).

I guess that even if this is not strictly necessary, I would still rather err on the side of being clearer and more verbose on what the behavior is for the benefit of package users that won't read the source code.

R/retry.R
asieira: I really think it's useful to output what the encountered error was, otherwise the information on what errors happened during retries will be completely lost. IMHO outputting this information will be invaluable to httr users to debug transient errors they are encountering. i.e., being able to tell a DNS resolution failure from a failed connection attempt:

```
> RETRY("GET", "http://invalidhostname/")
Request failed [Error in curl::curl_fetch_memory(url, handle = handle): Couldn't resolve host name
]. Retrying in 1 seconds...
```

Any alternative ways to output this, perhaps?

R/retry.R
hadley: I don't understand your comment. I just want you to replace "If missing or null", with "If null".

R/retry.R
hadley: Yes, it would be better to have a completely different print method in that case.

R/retry.R
asieira: I am so sorry about the confusion, I thought you wanted to remove the entire sentence that describes what the behavior is when the parameter was NULL or missing. My mistake. 🙇 

Will replace "if missing or null" for "if null" as request right now.

R/retry.R
asieira: Ok, updated the code to print the error condition on a separate line and include `ERROR` instead of the status code:
```
> RETRY("GET", "http://invalidhostname/")
Error in curl::curl_fetch_memory(url, handle = handle): Could not resolve host: invalidhostname
Request failed [ERROR]. Retrying in 2 seconds...
```

R/retry.R
```diff
@@ -25,33 +32,60 @@
...
+    !http_error(resp)
+  }
+}
+
+backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60, quiet = FALSE) {
   length <- ceiling(stats::runif(1, max = min(pause_cap, pause_base * (2 ^ i))))
   if (!quiet) {
-    message("Request failed [", status, "]. Retrying in ", length, " seconds...")
+    if (inherits(resp, "error")) {
+      error_description <- gsub("[\n\r]+$", "", as.character(resp))
```
hadley: I think `resp$message` is probably a better way of getting the error message.

asieira: That does work and makes the code a bit cleaner. But doing `as.character` returns more information, in particular where the error occurred:
```
> library(httr)
> e = tryCatch(GET("http://invalidhostname"), error = function(e) e)
> e$message
[1] "Could not resolve host: invalidhostname"
> as.character(e)
[1] "Error in curl::curl_fetch_memory(url, handle = handle): Could not resolve host: invalidhostname\n"
```

I vote for using `as.character` for the more complete information, even if that means the code is a little uglier. What do you think?

R/retry.R
```diff
@@ -1,7 +1,11 @@
 #' Retry a request until it succeeds.
 #'
-#' Safely retry a request until it succeeds (returns an HTTP status code
-#' below 400). It is designed to be kind to the server: after each failure
+#' Safely retry a request until it succeeds, as defined by the \code{terminate_on}
+#' parameter, which by default means a response for which \code{\link{http_error}()}
+#' is \code{FALSE}. Will also retry on error conditions raised by the underlying curl code,
+#' but if the last retry still raises one, \code{RETRY} will raise it again with
+#' \code{\link{stop}()} to maintain backwards compatibility.
```
hadley: Can you please remove "to maintain backwards compatibility"

R/retry.R
hadley: I forgot that `as.character.error()` existed, so I now think using `as.character()` is fine.

hadley:
Thanks for all your work on this!

asieira:
My pleasure. Thank *you* for all the wonderful packages! 🙇

mvkorpel:
Using the `terminate_on` parameter seems a bit unwieldy. Let's say I would like to stop retrying on code 403 (Forbidden). In my opinion, the intuitive way to do this is to specify `terminate_on = 403`. However, this does not have the desired effect, as even code 200 (OK) will then cause retries. To do the right thing, one must use `terminate_on = c(200, 403)`, or possibly something like `terminate_on = c(100:399, 403)` to account for all imaginable non-error codes.

The parameter works exactly as documented, but I think it would be great to have a user-friendly way to (by default?) include non-error codes (< 400) on top of the set given in `terminate_on`. Alternatively, a puzzled user might appreciate a note in the `?RETRY` document about the non-error codes.

Anyway, `terminate_on` is already useful, but thanks in advance for considering if a small change to the logic or a note in the manual would make sense!

jennybc:
@mvkorpel A comment on a merged PR is very low visibility, i.e. no one will necessarily see it when they come back to work on this package. I suggest you create a new issue with these observations about how to make RETRY more useful.

mvkorpel:
@jennybc Thanks for the comment. I will do that.

PR #463, opened 2017-07-30
Client credential v2

cderv:
This is the new PR following #388 discussion. It closes #384. 

+ Uses `client_credentials = TRUE` to activate use of client credential grant
+ Sets `code <- NULL`, `redirect_uri` to NULL (as not need for this grant) and set `grant_type ="client_credentials"` 
+ Tweaks documentation of `oauth_endpoint` for authorize url
+ Adds a demo file for YELP API to demonstrate the use of an API with client credential grant.
+ Adds a bullet to NEWS

I also generate new documentation. 

@hadley , what do you think of this new PR ?

PR #464, opened 2017-08-01
fix client credentials grant

cderv:
During resolution of merge conflict in #463, a line was lost. 

This PR is just there to add back this missing in order for #384 to work.

Sorry for that. When you resolve merge conflict directly on github, it is not easy to check. 

Will do this locally next time.


cderv:
and now a travis issue because of the new secrets vignette. 
Too bad it just missed this one line for client credential to work right away. 
We were so close !

hadley:
Thanks!

PR #467, opened 2017-08-01
Secrets feedback

jennybc:
I now see there's also a PR from @gaborcsardi, so there may be overlap?


jennybc:
Hmm let me figure out why the base of this PR is so wrong.

gaborcsardi:
There is some overlap, I'll be happy to fix my PR once this one is merged.

PR #471, opened 2017-08-02
minor changes/adds

hrbrmstr:
- 56-65 : Folks will likely either not know about or be thinking about file permissions, so this should work on macOS, Windows and Linux and is good practices in and out of multi-tenant environments
- 191 : `httr::content` for consistency
- 215 : Selective trust will give folks bad habits :-)
- 219 : You're likely trying to keep this simple for readers but `cyphr` does a ++gd job managing group secrets


hadley:
Thanks!

PR #472, opened 2017-08-04
Has re-added code for using value of "retry-after" header as waiting time if status_code = 429

nielsoledam:
I've reset my fork to the current master and re-added the code changes as discussed on https://github.com/r-lib/httr/pull/434.

The code now also respects  the value of pause_min.

### DIFF
--- a/R/retry.R
+++ b/R/retry.R
@@ -95,7 +95,9 @@ backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60,
     }
     if (status == 429) {
       retry_after <- resp$headers[["retry-after"]]
-      if (!is.null(retry_after)) length <- max(pause_min, as.numeric(retry_after))
+      if (!is.null(retry_after)) {
+        length <- max(pause_min, as.numeric(retry_after))
+      }
     }
     message(error_description, "Request failed [", status, "]. Retrying in ", round(length, 1), " seconds...")
   }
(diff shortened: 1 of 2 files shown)


R/retry.R
```diff
@@ -91,6 +91,10 @@ backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60,
       error_description <- ""
       status <- status_code(resp)
     }
+    if (status == 429) {
+      length_429 <- as.numeric(resp$headers[["retry-after"]])
+      if (!is.null(length_429) && length_429 >= pause_min) length <- length_429
```
hadley: I don't think the logic is quite right here - `length_429` will never be `NULL` because you coerce it to numeric

Inside the `if` statement, I'd prefer something like `length <- max(pause_min, length_429)`

R/retry.R
```diff
@@ -92,8 +94,8 @@ backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60,
       status <- status_code(resp)
     }
     if (status == 429) {
-      length_429 <- as.numeric(resp$headers[["retry-after"]])
-      if (!is.null(length_429) && length_429 >= pause_min) length <- length_429
+      retry_after <- resp$headers[["retry-after"]]
+      if (!is.null(retry_after)) length <- max(pause_min, as.numeric(retry_after))
```
hadley: Can you please put the condition code inside of `{}` (indented like the rest of the code)

nielsoledam: Was this what you meant?

hadley:
Perfect, thanks!

PR #479, opened 2017-09-09
update wording in docs

ChrisMuir:
Still needs to be roxygenised.

Edit to add: Just pushed the updated `httr-package.Rd` file, and updates to docs in `oauth-listener.R`


hadley:
Thanks!

PR #495, opened 2017-11-05
Normalize scopes in the hash; fixes #494

### LINKED ISSUE #494: Scopes should be sorted and name-stripped before hashing
This is obviously not urgent or it would have come up already. So maybe this is just for httr2 (#351).

But meaningless differences in the `scope` vector cause the hash to be different and therefore cause new tokens to be obtained when a suitable token is already in the cache.

Before hashing, I think `scope` should be

  * sorted
  * stripped of names

If you run the code below interactively, 3 tokens are eventually obtained, but really only 1 should be necessary.

```r
library(httr)

t1 <- oauth2.0_token(
  endpoint = oauth_endpoints("google"),
  app = googledrive:::oauth_app(),
  scope = c("https://www.googleapis.com/auth/drive", "email")
)
tokens <- readRDS(".httr-oauth")
names(tokens)
## there is one token

t2 <- oauth2.0_token(
  endpoint = oauth_endpoints("google"),
  app = googledrive:::oauth_app(),
  scope = c("https://www.googleapis.com/auth/drive", "email")
)
## no auth dance
tokens <- readRDS(".httr-oauth")
names(tokens)
## there is still the same, single token

## swap the order of the 2 scopes
t3 <- oauth2.0_token(
  endpoint = oauth_endpoints("google"),
  app = googledrive:::oauth_app(),
  scope = c("email", "https://www.googleapis.com/auth/drive")
)
## new oauth dance!
tokens <- readRDS(".httr-oauth")
names(tokens)
## there are two tokens

## scope vector has names
t4 <- oauth2.0_token(
  endpoint = oauth_endpoints("google"),
  app = googledrive:::oauth_app(),
  scope = c(a = "email", b = "https://www.googleapis.com/auth/drive")
)
## new oauth dance!
tokens <- readRDS(".httr-oauth")
names(tokens)
## there are three tokens
```

jennybc:
As I say in #494, it's not clear this is worth fiddling with now, but here's what I mean.


hadley:
Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jennybc:
Rebased and NEW bullet added.

PR #528, opened 2018-05-29
make cache_info() robust to flagless cache-control

mtmorgan:
- closes r-lib/httr#498


hadley:
Would you mind also adding a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks @mtmorgan!

PR #529, opened 2018-05-30
Fix unnecessary error on failure to parse headers$Date

stefanfritsch:
Hi,

when parsing the `headers$Date` in `request_perform` fails, e.g. because the response is not in `en_US` it results in an unhelpful prettyNum (or so) error.

I just split the if-else to replace the `date` with something valid if it failed to parse.

Ciao,
Stefan


hadley:
Could you please give me an example of the error that this resolves? I don't understand what you mean by `prettyNum`

hadley:
Oh I see this is a duplicate of #544, which explains the problem better.

PR #543, opened 2018-10-09
allow httr::POST to pass args to jsonlite::toJSON

drewabbot:
rationale: the way httr::POST wrap's jsonlite::toJSON is not quite general enough


hadley:
As discussed in the linked issue, I don't think this is a great approach (because passing `...` to multiple functions will lead to problems). Instead, just create the json yourself.

PR #544, opened 2018-10-10
Change the default failure in parse_http_date() to NA with class Date

shrektan:
The default value of `failure` argument in `parse_http_date()` should be NA with class Date.

Otherwise, `print.response()` may throw an unfriendly error.

On a Chinese-language Windows machine, I use plumber to build an REST API. The API returns the `header$date` in Chinese format by default, like "周三, 10 十月 2018 1:19:13 GMT". 

`parse_http_date()` will fail to parse it correctly and fall back to the default value of `failure``.

However, the default value of `failure` is set to `NA` (the logical value). Whenever I try to print the response, R will throw error saying:

`Error in prettyNum(.Internal(format(x, trim, digits, nsmall, width, 3L,  :   invalid 'trim' argument`

I believe the cause is https://github.com/r-lib/httr/blob/976289a3596dc01dc994f8fd743770a172bbecdb/R/response.r#L35

After this PR, the result can be printed successfully now.

```
Response [http://127.0.0.1:8001/reLoad]
  Date: NA
  Status: 200
  Content-Type: application/json
  Size: 2 B
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr 1.3.1.9000
 
+* The default value of `failure` argument in `parse_http_date()` is set to `structure(NA_real_, class = "Date")` so that the reponse with a "failure" date can be printed out correctly. (@shrektan, #544)
+
 # httr 1.3.1
 
 * Re-enable on-disk caching (accidentally disabled in #457) (#475)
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #548, opened 2018-11-24
add a demo for one-legged OAuth1 using Noun Project API 

### LINKED ISSUE #542: One legged OAuth1
Sorry but I can't seem to make httr work with the [Noun Project API](https://api.thenounproject.com/)). 
The site suggests a very simple access based just on the key and the secret. 

My R code is as follows:
```
 token <- oauth1.0_token(
      endpoint = oauth_endpoint(
          request = "http://api.thenounproject.com/",
          authorize = "http://api.thenounproject.com/icon/15",
          access = "http://api.thenounproject.com/icon/15"),
      app = oauth_app(
          appname = "nounprojectR",
          key = key,
          secret = secret))
```
It opens a browser window but gives a 
"Bad Request
Missing OAuth parameters"
response. 

This seems like it should be a really simple problem - like a gap between how httr and the API are handling the key and secret - perhaps API is not recognising the text of the key for some reason. 

I imagine I'm doing something really stupid but I'm really at my wits end trying to make this work and would love a bit of help if anybody recognises the problem. 

Thanks, 
Paul

cderv:
This closes #542 by providing a demo on using `oauth_signature` to use an OAuth1.0a One legged API.


cderv:
Unfortunately, I am not sure how to do for demo to provide Key and Secret. I think demo files are run at built time, or at least can be run using `demo()`. 

Currently I left the `Sys.getenv` but it won't work unless the user register himself and add the env to Renviron. It this ok ?

hadley:
Maybe use `rstudioapi::askForPassword()` instead of `Sys.getenv()`? Then it's one it's a bit more direct for the user.

You don't need to worry about the demos being automatically rerun.

demo/oauth1-nounproject.r
```diff
@@ -0,0 +1,44 @@
...
+# Noun Project has an API secured with OAuth 1.0a One-legged. A client key and a
+# secret must be used to sign requests when accessing the API. This is an
+# exemple using OAuth 1.0a One legged auth mechanism
+# http://oauthbible.com/#oauth-10a-one-legged
+
+# 1. Register an application to get required keys:
+# https://thenounproject.com/accounts/login/?next=/developers/apps/
+# Add the below  environment variable to your session
+# (helper function : usethis::edit_r_environ())
+nouns_app <- oauth_app("noun_project",
```
hadley: Would you mind using the standard tidyverse style here?

hadley:
Maybe use `rstudioapi::askForPassword()` instead of `Sys.getenv()`? Then it's one it's a bit more direct for the user.

You don't need to worry about the demos being automatically rerun.

hadley:
Actually, let me just finish this off since the changes are small and I need to practice my contributing back to PR skills

hadley:
Thanks @cderv!

cderv:
Thanks!
I really need to study and take care of the tidyverse style to stop having reviews about it 😉

Issue #493, opened 2017-10-27
support for OAuth2 in Yahoo API


ctrombley:
I was recently able to complete a successful auth flow against Yahoo's OAuth2 API with httr, but only after a few small changes:

1. The OOB flow [expects a request_uri value of `oob`](https://developer.yahoo.com/oauth2/guide/flows_authcode/#step-2-get-an-authorization-url-and-authorize-access) rather than [`urn:ietf:wg:oauth:2.0:oob`](https://github.com/r-lib/httr/blob/7150e8edd334d019788ee2d50ad831b48b9c6df0/R/oauth-init.R#L85).
1. The base url for Yahoo's OAuth2 endpoints had to be changed from [`https://api.login.yahoo.com/oauth/v2`](https://github.com/r-lib/httr/blob/master/R/oauth-endpoint.r#L89) to   [`https://api.login.yahoo.com/oauth2`](https://developer.yahoo.com/oauth2/guide/flows_authcode/#step-2-get-an-authorization-url-and-authorize-access).

Is there any desire to add either of these changes into the package?  I can take a stab at a PR if so.  For the first issue, it would be simple to add an override for the OOB string, with the option scoped either globally or to the app.  For the second, the existing endpoint could be updated or separate endpoints could be provided for OAuth1 and OAuth2.

Thoughts?

Tectra:
@ctrombley would you be willing to sure your code? been struggling with getting oauth2 working.

ctrombley:
@Tectra sure thing.  Keep in mind I am running an OOB OAuth flow, so I have provisioned a Desktop application in Yahoo.  Here's a gist:

https://gist.github.com/ctrombley/296c2d9eb2738642b953debfc7404b12

It requires my httr fork available here:
https://github.com/ctrombley/httr

Also worth noting that I am creating my own `oauth_app` rather than using `oauth_endpoints("yahoo")`. Here's the commit in my fork that allows overriding the `redirect_url` at the app level:
https://github.com/ctrombley/httr/commit/86b409d5036cf143b113290c2faf165a4aefd099

hadley:
@ctrombley are you still interested in doing a PR? I'm happy to implement myself, but if you'd like to take a shot I'd be happy to review your code. Please let me know either way.

If so, the main challenge is figuring out how to specify the oob value. For now, since we'll reconsider this entire API in httr2, I'd recommend adding a new parameter, `oob_value`

hadley:
(And also needs an update of the demo)

[...]

PR #550, opened 2018-11-27
#493 Add support for custom oob redirect values

ctrombley:
cc @hadley 

For #493. This PR adds support for specifying a custom OOB parameter value when retrieving an authorization URL.  The Yahoo docs here show that their implementation requires a `redirect_uri` value of "oob":

https://developer.yahoo.com/oauth2/guide/flows_authcode/#step-2-get-an-authorization-url-and-authorize-access=

This PR also includes the following:
* An updated Yahoo OAuth2 demo
* An updated Yahoo base API URL
* Updated docs


R/oauth-init.R
```diff
@@ -72,17 +74,20 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL,
...
                           client_credentials = FALSE
                          ) {
 
   scope <- check_scope(scope)
-  use_oob <- check_oob(use_oob)
-
+  use_oob <- check_oob(use_oob, oob_value = NULL)
   if (use_oob) {
-    redirect_uri <- "urn:ietf:wg:oauth:2.0:oob"
+    redirect_uri <- if(!is.null(oob_value))
```
hadley: Please either write on one line, or use `{}`

R/oauth-init.R
```diff
@@ -57,6 +57,8 @@ init_oauth1.0 <- function(endpoint, app, permission = NULL,
 #'   Otherwise, provide a URL to the user and prompt for a validation
 #'   code. Defaults to the of the \code{"httr_oob_default"} default,
 #'   or \code{TRUE} if \code{httpuv} is not installed.
+#' @param oob_value if provided, is used as a custom oob callback value.
```
hadley: Can you please make this description a bit more precise? It's not a callback; it's passed as a redirect parameter.

tests/testthat/test-oauth.R
```diff
@@ -53,6 +53,30 @@ test_that("oauth_encode1 works", {
   expect_equal(orig_string, restored_string)
 })
 
+test_that("custom oob callback values can be provided", {
```
hadley: TBH I don't think this is worth it; in httr2 we'll refactor to make this sort of testing easier, but for now I think it's better to omit as the test will be fragile.

ctrombley: Removed.

R/oauth-init.R
ctrombley: Done.

R/oauth-init.R
ctrombley: Done.

R/oauth-init.R
```diff
@@ -72,17 +75,17 @@ init_oauth2.0 <- function(endpoint, app, scope = NULL,
...
                           client_credentials = FALSE
                          ) {
 
   scope <- check_scope(scope)
-  use_oob <- check_oob(use_oob)
-
+  use_oob <- check_oob(use_oob, oob_value)
   if (use_oob) {
-    redirect_uri <- "urn:ietf:wg:oauth:2.0:oob"
+    redirect_uri <- ifelse(!is.null(oob_value), oob_value, "urn:ietf:wg:oauth:2.0:oob")
```
hadley:
```suggestion
    redirect_uri <- if (!is.null(oob_value)) oob_value else "urn:ietf:wg:oauth:2.0:oob"
```

`ifelse()` is specifically for creating vectors; you can just use an inline `if` here

ctrombley:
Done. I updated the package to 1.8.2.  Let me know if there’s anything else I can help with!

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: httr
 Title: Tools for Working with URLs and HTTP
-Version: 1.3.1.9000
+Version: 1.3.2
```
hadley: Please revert this change - the development version always ends in .9000 (and it's my job to increment when necessary)

ctrombley: Done.  I also moved my bullets in NEWS.md under `1.3.1.9000`.

hadley:
Thanks @ctrombley!

Issue #561, opened 2019-01-03
Submitting an explicitly NULL parameter via httr


peterhurford:
Is there a way to submit NULL as a JSON parameter via httr?

When I do `httr::POST("https://httpbin.org/post", body = list(a = 1, b = NULL), httr::verbose(), encode = "json”)`, I see in the output that `b = NULL` gets dropped from the payload somewhere, even though `list(a = 1, b = NULL)` is a valid R list that has an existent b value set to `NULL`.

In Python's requests library, submitting a None parameter is permitted and I'm using an API that depends on that behavior (maybe not the smartest design choice, but it's what I have to live with). Is there any way for `httr` or `RCurl` or something to interface with this?

cderv:
here are some ways to make it works

You can use a raw body for that

``` r
httr::POST("https://httpbin.org/post", body = '{"a":1,"b":{}}', encode = "raw")
#> Response [https://httpbin.org/post]
#>   Date: 2019-01-03 20:53
#>   Status: 200
#>   Content-Type: application/json
#>   Size: 464 B
#> {
#>   "args": {}, 
#>   "data": "{\"a\":1,\"b\":{}}", 
#>   "files": {}, 
#>   "form": {}, 
#>   "headers": {
#>     "Accept": "application/json, text/xml, application/xml, */*", 
#>     "Accept-Encoding": "gzip, deflate", 
#>     "Connection": "close", 
#>     "Content-Length": "14", 
#> ...
```

or build the body directly using `jsonlite`

``` r
json_body <- jsonlite::toJSON(list(a = 1, b = NULL), auto_unbox = TRUE)
httr::POST("https://httpbin.org/post", body = json_body, encode = "raw")
#> Response [https://httpbin.org/post]
#>   Date: 2019-01-03 20:53
#>   Status: 200
#>   Content-Type: application/json
#>   Size: 464 B
#> {
#>   "args": {}, 
#>   "data": "{\"a\":1,\"b\":{}}", 
#>   "files": {}, 
#>   "form": {}, 
#>   "headers": {
#>     "Accept": "application/json, text/xml, application/xml, */*", 
#>     "Accept-Encoding": "gzip, deflate", 
#>     "Connection": "close", 
#>     "Content-Length": "14", 
#> ...
```

<sup>Created on 2019-01-03 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

In fact, it think it is how httr work because if `body` is a list, 
[...]

Issue #539, opened 2018-09-21
POST could take toJSON parameters


viddagrava:
If `encoding="json"`, it would be really helpful if `jsonlite::toJSON` parameters could be provided to the `POST` function (or also in the other request types).

Right now only `auto_unbox = TRUE` is applied, however, it could be quite flexible if we could provide the rest of the parameters as well.

I have a use case where I have to send geo-coordinates to an API within a json object. In order to overcome the `digits=4` default parameter of `toJSON`, I have to manually encode my object and send the request as `"raw"`. This is my current workaround.

Any feedback?

shrektan:
I'm looking for the answer, too. After some searching, this, https://github.com/r-lib/httr/pull/364 and https://github.com/r-lib/httr/issues/455 , I decide to use raw as the workaround...

hadley:
I don't think there's any way to make this work with the current httr API, but it's definitely something we'll think about in httr2.  For now, as @shrektan suggests, just convert to json yourself

PR #563, opened 2019-01-04
Add an example about empty element in body for POST

cderv:
Following discussion in #561, I added an example in POST for this specific case. This show how to use `encode = "raw"` for providing json when the automatic `encode = "json"` does not suit.

This cover also #539 question.


hadley:
Thanks!

hadley:
Thanks!

PR #570, opened 2019-02-07
Support for additional claims in the JWT tokens, needed for Google's Identity-Aware Proxy

kalugny:
Hi and thanks for an excellent package!

We're using Google Cloud's Identity-Aware Proxy (IAP) for which `httr` is _almost_ compatible.
The only discrepancy is that IAP requires an additional "claim" on the JWT.

This pull requests:
1.  Adds support for additional claims on the JWT, which you can pass as additional parameters to `oauth_service_token`
2. Supports getting the authentication header from the `id_token` if `access_token` is not set (which is the case for IAP tokens)

Using this for IAP is now very simple:
```r
token <- oauth_service_token(
  endpoint = oauth_endpoints("google"),
  secrets = jsonlite::fromJSON('/path/to/credentials.json'),
  target_audience = iap_oauth_client_id
)
```


kalugny:
This is the google docs for it:
https://cloud.google.com/iap/docs/authentication-howto#authenticating_from_a_service_account

kalugny:
@hadley any comments? Should I change something?
It's a simple extension that would support more use-cases.

R/oauth-server-side.R
```diff
@@ -42,6 +43,7 @@ init_oauth_service_account <- function(secrets, scope = NULL, sub = NULL) {
 #'   00:00:00 UTC, January 1, 1970. This value has a maximum of 1 hour from
 #'   the issued time.
 #' @param duration Duration of token, in seconds.
+#' @param ... any additional claims for the token.
```
hadley, line 46:
```suggestion
#' @param ... Any additional claims for the token.
```

R/oauth-token.r
```diff
@@ -301,6 +301,7 @@ Token2.0 <- R6::R6Class("Token2.0", inherit = Token, list(
 #' @inheritParams oauth2.0_token
 #' @inheritParams jwt_signature
 #' @param secrets Secrets loaded from JSON file, downloaded from console.
+#' @param ... any additional claims for the token
```
hadley, line 304: I think you can remove this line since we're already inheriting docs from `jwt_signature`?

R/oauth-token.r
```diff
@@ -311,7 +312,7 @@ Token2.0 <- R6::R6Class("Token2.0", inherit = Token, list(
 #'
 #' token <- oauth_service_token(endpoint, secrets, scope)
 #' }
-oauth_service_token <- function(endpoint, secrets, scope = NULL, sub = NULL) {
+oauth_service_token <- function(endpoint, secrets, scope = NULL, sub = NULL, ...) {
```
hadley, line 315: Using `...` here seems a bit broad. I'd think it should be `claims` or similar?

hadley:

PR #572, opened 2019-02-12
added handling of simpleError in http_error (fixes #569)

### LINKED ISSUE #569: RETRY error message
I was wondering if the message for the invalidhostname example shown in the `RETRY` documentation is the intended result:

https://httr.r-lib.org/reference/RETRY.html

```r
RETRY("GET", "http://invalidhostname/")
#> Error in UseMethod("http_error"): no applicable method for 'http_error' applied to an object of class "c('simpleError', 'error', 'condition')"
```

I had assumed it would be something more like the `GET` message:
```r
GET("http://invalidhostname/")
Error in curl::curl_fetch_memory(url, handle = handle) : 
  Could not resolve host: invalidhostname
```

jameslamb:
This PR attempts to address #569 so that failed requests will return a more helpful error message. I hope you'll consider it!

### DIFF
--- a/R/response-status.r
+++ b/R/response-status.r
@@ -180,6 +180,11 @@ http_error.integer <- function(x, ...) {
 }
 
 #' @export
+http_error.simpleError <- function(x, ...){
+    FALSE
+}
+
+#' @export
 #' @rdname http_error
 #' @usage NULL
 url_success <- function(x, ...) {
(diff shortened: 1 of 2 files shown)


austin3dickey:
@hadley Any update on this? I just ran into this issue as well.

jameslamb:
@hadley could I get a review on this?

hadley:
That doesn't look like the right fix to me. I'm not planning on working on httr in the near future, but I think the fix should come through rearranging the `if` conditions in `httr:::retry_should_terminate`

PR #581, opened 2019-04-01
Fix RETRY() error-handling

### LINKED ISSUE #569: RETRY error message

### LINKED ISSUE #574: RETRY errors on curl timeout
I recently started trying to use httr for RETRY because I've been getting intermittent errors in the middle of a long set of remote API calls. However, it seems like when CURL times out- and doesn't provide an HTTP status code- RETRY dies. I would expect that RETRY would retry even in these cases. Since the documentation doesn't mention this kind of timeout, I'm not sure if this is a bug report or a feature request.

Simple reproducible example-
This dies:
`RETRY("GET", "http://httpbin.org/delay/3", timeout(1))`
`#Error in UseMethod("http_error") : 
  no applicable method for 'http_error' applied to an object of class "c('simpleError', 'error', 'condition')"`

This is fine:
`RETRY("GET", "http://httpbin.org/delay/3", timeout(4))`

R version 3.5.1 (2018-07-02)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200
httr_1.4.0.9000

austin3dickey:
This PR fixes a few issues where an unsuccessful `request_perform()` during a `RETRY()` will prematurely fatal with the following error message:
```
Error in UseMethod("http_error") : no applicable method for 'http_error' applied to an object of class "c('simpleError', 'error', 'condition')"
```
Now, `RETRY()` will successfully retry in those cases. This fix was @hadley 's [suggestion](https://github.com/r-lib/httr/pull/572#issuecomment-477588255).

Fixes #569 
Fixes #574

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # httr (development version)
 
+* `RETRY()` now throws the correct error message if an error occurs during the 
+  request (@austin3dickey, #581).
+
 # httr 1.4.0
 
 ## OAuth
--- a/R/retry.R
+++ b/R/retry.R
@@ -77,12 +77,12 @@ RETRY <- function(verb, url = NULL, config = list(), ...,
 }
 
 retry_should_terminate <- function(i, times, resp, terminate_on, terminate_on_success) {
-  if (terminate_on_success && !http_error(resp)) {
-    TRUE
-  } else if (i >= times) {
+  if (i >= times) {
     TRUE
   } else if (inherits(resp, "error")) {
     FALSE
+  } else if (terminate_on_success && !http_error(resp)) {
+    TRUE
   } else if (!is.null(terminate_on)) {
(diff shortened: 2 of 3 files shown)


hadley:
Thanks!

PR #583, opened 2019-04-01
Uppercase methods

### LINKED ISSUE #571: Automatically uppercase methods
Would it be more user-friendly to automatically uppercase `GET, POST, PUT, PATCH` etc? The [http]( https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html) specification has the verbs in all caps and many (all?) webservers reject lowercase. 

An example:

```
httr::VERB('get', 'google.com')
```
```
Response [http://google.com/]
  Date: 2019-02-08 20:38
  Status: 405
  Content-Type: text/html; charset=UTF-8
  Size: 1.59 kB
<!DOCTYPE html>
<html lang=en>
  <meta charset=utf-8>
  <meta name=viewport content="initial-scale=1, minimum-scale=1, width=device-width">
  <title>Error 405 (Method Not Allowed)!!1</title>
  <style>
    *{margin:0;padding:0}html,code{font:15px/22px arial,sans-serif}html{background:#fff;color:#222;padding:15px}body{margin:7% auto 0;max-...
  </style>
  <a href=//www.google.com/><span id=logo aria-label=Google></span></a>
  <p><b>405.</b> <ins>That’s an error.</ins>
```
Whereas, the upper case version returns a 200 successfully:
```
httr::VERB("GET", "google.com")
```
```
Response [http://www.google.com/]
  Date: 2019-02-08 20:41
  Status: 200
  Content-Type: text/html; charset=ISO-8859-1
  Size: 12.6 kB
<!doctype html><html itemscope="" itemtype="http://schema.org/WebPage" lang="en"><head><meta content="Search the world's information, incl...
</style><style>body,td,a,p,.h{font-family:arial,sans-serif}body{margin:0;overflow-y:scroll}#gog{padding:3px 8px 0}td{line-height:.8em}.gac...
if (!iesg){document.f&&document.f.q.focus();document.gbqf&&document.gbqf.q.focus();}
}
})();</script><div id="mngb"> <div id=gbar><nobr><b class=gb1>Search</b> 
```

A reason to do so is that the openAPI swagger specification specifies verbs in all lower case [see here](https://github.com/swagger-api/swagger-parser/issues/550). Translating a swagger spec directly to API calls requires a user to upper case the verbs.

I realize that this is a speculative and potentially breaking change in behavior. But did want to at least raise the issue.

patr1ckm:
This uses uppercase methods everywhere by default, for use primarily in `VERB` and `RETRY`. 

This approach does a bit of extra work but was easier to test. An alternative is to implement the change only for `VERB` and `RETRY`, which I can do if you prefer.

Note that the curl options in `request_prepare` will now be set for formerly lowercase `get` and `post` calls, which were handled as custom calls previously. 
```
    GET = req$options$httpget <- TRUE,
    POST = req$options$post <- TRUE,
```

The result is that 
```
httr::VERB('get', 'google.com')
```
Now returns a 200.

Fixes #571 .

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 * `RETRY()` now throws the correct error message if an error occurs during the 
   request (@austin3dickey, #581).
 
+* `VERB()` and `RETRY()` now automatically uppercase methods (@patr1ckm, #571).
+
 # httr 1.4.0
 
 ## OAuth
--- a/R/http-verb.R
+++ b/R/http-verb.R
@@ -14,13 +14,13 @@
 #' )
 #' stop_for_status(r)
 #' content(r)
-#'
+#' 
 #' VERB("POST", url = "http://httpbin.org/post")
 #' VERB("POST", url = "http://httpbin.org/post", body = "foobar")
 VERB <- function(verb, url = NULL, config = list(), ...,
                  body = NULL, encode = c("multipart", "form", "json", "raw"),
                  handle = NULL) {
   hu <- handle_url(handle, url, ...)
-  req <- request_build(toupper(verb), hu$url, body_config(body, match.arg(encode)), config, ...)
+  req <- request_build(verb, hu$url, body_config(body, match.arg(encode)), config, ...)
   request_perform(req, hu$handle$handle)
 }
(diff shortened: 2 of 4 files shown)


hadley:
Thanks!

Issue #567, opened 2019-01-21
Error in http_statuses[[as.character(status)]] :   subscript out of bounds


Javdat:
Sometimes I am getting an error `Error in http_statuses[[as.character(status)]] :   subscript out of bounds` while calling `s3saveRDS()` in `aws.s3`. As far as I could pinpoint, the issue originates from `httr`, more specifically in `http_status`.

This is the part of the code that is (I think) is problematic:

```
  status <- status_code(x)

  status_desc <- http_statuses[[as.character(status)]]
  if (is.na(status_desc)) {
    stop("Unknown http status code: ", status, call. = FALSE)
  }
```

This assumes that `http_statuses[[as.character(status)]]` will return `NA` when it is faced with a status code not present in `http_statuses`, and thus will be handed with: "Unknown http status code: ".

However, in reality, it seems `http_statuses[[as.character(status)]]` will just throw `subscript out of bounds` error.

Reproducible(ish) example:

Status present in `http_statuses`, works as expected:
```
> status <- status_code(100)
> status_desc <- httr:::http_statuses[[as.character(status)]]
> status_desc
[1] "Continue"
```
Unknown status, fails before reaching "Unknown http status code: " handler:

```
> status <- status_code(1000)
> status_desc <- httr:::http_statuses[[as.character(status)]]
Error in httr:::http_statuses[[as.character(status)]] : 
  subscript out of bounds
> status_desc
Error: object 'status_desc' not found
```

So, I think either `httr:::http_statuses[[as.character(status)]]` has to be called safely and error handled as NA, or indexing should happen one level only.

```
> status <- status_code(1000)
> status_desc <- httr:::http_statuses[as.character(status)] # Notice [] instead of [[]]
> status_desc
<NA> 
  NA 
> as.character(status_desc) # remove variable name
[1] NA
```

sheffe:
Found this existing issue after lots of similar areas. For what it's worth, I make ~100k read/write calls through `aws.s3` every week, and I see this error on something like 1 in 200 of those calls. I've done some limited diagnostic checks to see whether this error happens more often on specific types of calls, payload sizes, etc., but it seems pretty random. 

(I haven't been able to create a full reprex without doing something objectively dumb, like "write 100k tiny RDS files with `aws.s3::s3saveRDS` and see what happens." The example above is a nice demo of what happens.)

[...]

PR #585, opened 2019-04-08
FIX unknown status code (#567)

Javdat:
Fixes issue (https://github.com/r-lib/httr/issues/567) with unknown status code. Slightly modified solution from original offered by @hadley.

### DIFF
--- a/R/response-status.r
+++ b/R/response-status.r
@@ -46,10 +46,11 @@ status_code.numeric <- function(x) x
 http_status <- function(x) {
   status <- status_code(x)
 
-  status_desc <- http_statuses[[as.character(status)]]
-  if (is.na(status_desc)) {
+  if (!status %in% names(http_statuses)) {
     stop("Unknown http status code: ", status, call. = FALSE)
   }
+
+  status_desc <- http_statuses[[as.character(status)]]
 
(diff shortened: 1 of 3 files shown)


R/response-status.r
```diff
@@ -46,11 +46,12 @@ status_code.numeric <- function(x) x
 http_status <- function(x) {
   status <- status_code(x)
 
-  status_desc <- http_statuses[[as.character(status)]]
-  if (is.na(status_desc)) {
+  if (isFALSE(status %in% names(http_statuses))) {
```
hadley: `isFALSE()` isn't a good approach here because it's only available in recent versions of R, and it's strictness isn't needed as we know that `%in%` always returns a logical vector with the same length as the LHS.

Javdat: `isFALSE` was here mostly to handle cases if status is `NA` for some reason. But apparently, `%in%` already handles it as well. Force changed the code.

Javdat:
Updated.

hadley:
For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

PR #586, opened 2019-04-09
Fix flaky RETRY unit test

jameslamb:
I noticed that recent builds of `httr` are failing (e.g. https://travis-ci.org/r-lib/httr/jobs/517331103) with an error like:

image

I could not reproduce this problem on my Mac tonight, which suggests to me that it has something to do with different underlying versions of either the `curl` R package or `libcurl` in the CI environment vs. my local dev environment.

In this PR, I propose making the `regexp` in that test less specific. I do still think it's important to have _some_ `regexp` there to be sure you got the actual curl error surfaced (since that was the intent of #581).

Thanks for considering this. I think it will unblock #582 , #583 , and #585

### DIFF
--- a/tests/testthat/test-retry.R
+++ b/tests/testthat/test-retry.R
@@ -17,6 +17,6 @@ test_that("sucessful requests terminate when terminate_on_success is true", {
 test_that("if request_perform() throws an error, RETRY passes it on", {
   expect_error(
     RETRY("POST", "http://98d90a2a254647889e2e4c236fb576cd.com", times = 1),
-    regexp = "Could not resolve host"
+    regexp = "resolve host"
   )
 })


jameslamb:
The build errors look unrelated to my changes:

image

I saw a lot of this type of error with the breaking changes in `devtools 2.0.0`. I don't see anything obvious in the `.travis.yml`...is the issue present on all `language: r` builds on Travis? 😬

jameslamb:
I ended up force-pushing with 0 changes to get Travis to kick off again

jameslamb:
This worked! Think whatever was going on with the build environment for R projects on Travis was resolved.

@hadley at your earliest convenience, could I get a review?

hadley:
Thanks. But please don't continue to ping me directly, especially for something as minor as this.

jameslamb:
Sure, sorry. Understood.

PR #595, opened 2019-05-17
Add pkgdown URL and update config.

batpigandme:
* Add pkgdown URL to `DESCRIPTION`.
* Add template config to `_pkgdown.yml`

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,5 +1,12 @@
 url: https://httr.r-lib.org
 
+template:
+  params:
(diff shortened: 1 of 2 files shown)


_pkgdown.yml
```diff
@@ -1,5 +1,12 @@
 url: https://httr.r-lib.org
 
+template:
+  params:
+    docsearch:
```
hadley: I think you can leave the search bit off

hadley:
Thanks!

PR #600, opened 2019-06-30
find_cert_bundle() check for CURL_CA_BUNDLE on all platforms

rnorberg:
Is there a reason this is only checked for on Windows? Linux users might also want to point the package to their system CA bundle (e.g. /etc/ssl/certs/ca-certificates.crt). I changed find_cert_bundle() to look for the CURL_CA_BUNDLE environment variable on all platforms. I also added a check to make sure that if CURL_CA_BUNDLE is set, it's set to a file that exists.

### DIFF
--- a/R/utils.r
+++ b/R/utils.r
@@ -74,18 +74,23 @@ keep_last <- function(...) {
 }
 
 find_cert_bundle <- function() {
-  if (.Platform$OS.type != "windows") {
-    return()
-  }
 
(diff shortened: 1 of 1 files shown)


jeroen:
There is no need for this. libcurl will automatically pick up `CURL_CA_BUNDLE` if set, there is no need to do this manually.

Also libcurl on Windows and MacOS do not need a cert bundle at all because they use the system native SSL backend.

PR #614, opened 2019-09-09
FIX: Retry-After header when quiet=TRUE

### LINKED ISSUE #611: RETRY() doesn't consult the Retry-After header if called with `quiet = TRUE`
The API I work with sends a `Retry-After` header when rate limiting. `httr::RETRY` respects only `retry-after` ([line](https://github.com/r-lib/httr/blob/0dc75595627e9c82ac6950320f4ea8e767bae378/R/retry.R#L105)). It seems that at least one definition of [the protocol](https://tools.ietf.org/html/rfc7231#section-7.1.3l) specifies `Retry-After` as the standard. Should `RETRY` automatically respect both cases by default?

patr1ckm:
`RETRY` now respects `Retry-After` and `retry-after` headers. `Retry-After` is the official standard.  Also fixes a previous bug that ignores the `Retry-After` if `quiet=TRUE`.

Fixes #611 .

### DIFF
--- a/R/retry.R
+++ b/R/retry.R
@@ -93,21 +93,23 @@ retry_should_terminate <- function(i, times, resp, terminate_on, terminate_on_su
 backoff_full_jitter <- function(i, resp, pause_base = 1, pause_cap = 60,
                                 pause_min = 1, quiet = FALSE) {
   length <- max(pause_min, stats::runif(1, max = min(pause_cap, pause_base * (2^i))))
+  if (inherits(resp, "error")) {
+    error_description <- gsub("[\n\r]*$", "\n", as.character(resp))
+    status <- "ERROR"
+  } else {
+    error_description <- ""
+    status <- status_code(resp)
+  }
+  if (status == 429) {
+    retry_after <- resp$headers[["retry-after"]] %||%
+      resp$headers[["Retry-After"]]
+    if (!is.null(retry_after)) {
+      length <- max(pause_min, as.numeric(retry_after))
+    }
+  }
   if (!quiet) {
-    if (inherits(resp, "error")) {
-      error_description <- gsub("[\n\r]*$", "\n", as.character(resp))
-      status <- "ERROR"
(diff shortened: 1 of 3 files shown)


hadley:
I think this needs to be fixed further upstream — something is missing the `insensitive` class that would make this work automatically.

R/retry.R
```diff
@@ -93,21 +93,23 @@ retry_should_terminate <- function(i, times, resp, terminate_on, terminate_on_su
...
+  if (inherits(resp, "error")) {
+    error_description <- gsub("[\n\r]*$", "\n", as.character(resp))
+    status <- "ERROR"
+  } else {
+    error_description <- ""
+    status <- status_code(resp)
+  }
+  if (status == 429) {
+    retry_after <- resp$headers[["retry-after"]] %||%
+      resp$headers[["Retry-After"]]
```
jennybc, line 105: I'm all for the change to make this case-insensitive. But I think we might as well go all the way and use an approach that would match against, e.g., `Retry-after`, as well. Not that I expect that specifically, more on principle.

jennybc: Reading more, perhaps this is encompassed by @hadley's comment, in fact.

jennybc: Maybe what you need to do @patr1ckm is search for other uses of `insensitive()` here and incorporate that into your PR.

patr1ckm:
Yes, I think I was wrong about insensitivity. The header should be all lower case coming from:
https://github.com/r-lib/httr/blob/af25ebd0e3b72d2dc6e1423242b94efc25bc97cc/R/headers.r#L135

I do think the change about respecting the header regardless of `quiet=TRUE` still holds. I'll get this cleaned up!

jennybc:
Reading even more, yeah either the insensitive class isn't behaving as intended or the specific call that checks the header is unintentionally bypassing the functionality of the class and is, therefore, getting case sensitive lookup.

PR #615, opened 2019-09-18
URL's scheme parser is compliant with RFC 3986

ymarcon:
According to [RFC 3986 #3.1](https://tools.ietf.org/html/rfc3986#section-3.1) the scheme syntax is: 

`scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`

The `parse_url()` function is modfified accordingly.

Note that RFC 3986 has made obsolete RFC 1808 (referred by the original parse_url implementation).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # httr (development version)
 
+* `parse_url()` now refers to RFC3986 for the parsing of the URL's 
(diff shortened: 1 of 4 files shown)


hadley:
Can you please also update the docs to point to the new RFC?

ymarcon:
Documentation updated, please review.

hadley:
Thanks!

PR #620, opened 2019-10-15
Fix to missing scheme handling (defaults to http)

jonathon-love:
Fixes r-lib/httr#619
Fixes r-lib/httr#607

### DIFF
--- a/R/request.R
+++ b/R/request.R
@@ -157,6 +157,9 @@ request_perform <- function(req, handle, refresh = TRUE) {
   }
 
   url_scheme <- parse_url(resp$url)$scheme
+  if (is.null(url_scheme))
+    url_scheme <- "http"
+
   is_http <- tolower(url_scheme) %in% c("http", "https")
   if (is_http) {
     all_headers <- parse_http_headers(resp$headers)
--- a/R/request.R
+++ b/R/request.R
@@ -157,8 +157,9 @@ request_perform <- function(req, handle, refresh = TRUE) {
   }
 
   url_scheme <- parse_url(resp$url)$scheme
-  if (is.null(url_scheme))
+  if (is.null(url_scheme)) {
     url_scheme <- "http"
+  }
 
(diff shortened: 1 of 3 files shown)


cderv:
Thanks ! 
Two questions I have on this: 

I am just wondering if we should assume that any parse url without scheme should be `http`. why not `https` ? 

Also, I wonder if this type of fix (if scheme is null, assume http) should be better in `request_perform` directly. 
I see `parse_url` as a utility function that helps get the pieces that compose a url. Currently, It could be used to know if a url as a scheme or not by testing if it returns null - with this fix, it can't. Do we want this change in `httr::parse_url`... 🤔 

Just sharing thoughts to help decide what is best.

jonathon-love:
> I am just wondering if we should assume that any parse url without scheme should be `http`. why not `https` ?

the standard behaviour of web-browsers et al. is to default to http when no scheme is provided. and it would be consistent with the earlier versions of `httr` built against curl 3. (additionally, most https sites issue a redirect from `http://` to `https://` anyway)

> Also, I wonder if this type of fix (if scheme is null, assume http) should be better in request_perform directly.

oh yeah, that's a fair point. i'll make that change.

with thanks

R/request.R
```diff
@@ -157,6 +157,9 @@ request_perform <- function(req, handle, refresh = TRUE) {
   }
 
   url_scheme <- parse_url(resp$url)$scheme
+  if (is.null(url_scheme))
```
hadley, line 160: Can you please use `{}` as in https://style.tidyverse.org/syntax.html#indenting?

petermeissner:
I have an [issue](https://github.com/ropensci/robotstxt/issues/50) reported in {robotstxt} directly related to this PR and also solved by this PR.

Will this PR be included in httr at some time in the future?

hadley:
We don't have any plans to spin off a httr release at the moment.

petermeissner:
Thanks, for the heads up.

hadley:

PR #634, opened 2020-01-17
Enable override the is_interactive function.

jobdiogenes:
In some  cases checking if is **interactive** mode are not as we need.
Since is_interactive parameter are deprecated and the is_interactive function enable to mocked  the native interactive function.
This pull request  add  just two lines to **export**  _is_interactive_ function that is in 
oauth-init.R,  **enabling** developers to **override** _is_interactive_.

I use this to help me use R and Google Drive / Sheets in Google Colaboratory, which are also useful to many users.
But its also useful to others cases and tests.


jennybc:
I suspect a better eventual solution is for httr to use `rlang::is_interactive()`, which did not exist when httr first appeared.

https://rlang.r-lib.org/reference/is_interactive.html

That affords various options for temporarily overriding the inferred interactive state. It would be odd for httr itself to export such a function.

As maintainer of the packages you are probably using for "R and Google Drive / Sheets in Google Colaboratory", I'd be curious to hear more. Now that I think about it, I suspect this is related to https://github.com/tidyverse/googledrive/issues/284, which I think is a more profound problem than just overriding `interactive()`. But I'd be curious to hear more about your workarounds, over in that issue.

jobdiogenes:
@jennybc using this workaround I could successfully use googledrive and googlesheets4 to access files and sheets. Here is [Workbook example]. (https://gist.github.com/jobdiogenes/8f0e8c82aaa6f7f164d667be4b69637f)

Yes, changing to:
```R
is_interactive <- function() rlang::is_interactive()

also solve  with use of:  **options(rlang_interactive=TRUE)**
which its more simple and light. 

another hack would be verify if is in colab. which have the advantage  that workbooks could works seamlessly without changes. but its a terrible hack.  
```R
is_interactive <- function() { 
  if (file.exists("/usr/local/lib/python3.6/dist-packages/google/colab/_ipython.py")) {
     return(TRUE) 
  } else { 
    return(interactive())
  }
}

- Yes please. make that change that you talk . that will solve the problem.

Thanks

**Comment by nacnudus** [comment]
@jobdiogenes Do you have an example colab notebook where you can connect to Google Drive?  I have tried [here](https://colab.research.google.com/drive/1GVgDcuh_1gNhRae1bTI3ZelOYllg8Egu) using your fork, but it still fails.

```r
devtools::install_github("jobdiogenes/httr")
devtools::install_github("tidyverse/googledrive")
library(googledrive)
options(rlang_interactive = TRUE)
drive_auth(use_oob = TRUE)
```

```
Is it OK to cache OAuth access credentials in the folder '/root/.R/gargle/gargle-oauth' between R sessions?
Error: Can't get Google credentials.
Are you running googledrive in a non-interactive session? Consider:
  * `drive_deauth()` to prevent the attempt to get credentials.
  * Call `drive_auth()` directly with all necessary specifics.
  * Read more in: https://gargle.r-lib.org/articles/non-interactive-auth.html
Traceback:

1. drive_auth(use_oob = TRUE)
2. stop("Can't get Google credentials.\n", "Are you running googledrive in a non-interactive session? Consider:\n", 
 .     "  * `drive_deauth()` to prevent the attempt to get credentials.\n", 
 .     "  * Call `drive_auth()` directly with all necessary specifics.\n", 
 .     "  * Read more in: https://gargle.r-lib.org/articles/non-interactive-auth.html", 
 .     call. = FALSE)
```

jobdiogenes:
@nacnudus Yes,  I did some more tests. 
And make a Notebook that are working, for both in colab or external.  and publish as Gist.

[ demo-of-use-google-drive-sheets-in-r-with-google-colaboratory.ipynb](https://gist.github.com/jobdiogenes/235620928c84e604c6e56211ccf681f0)

nacnudus:
Thank you @jobdiogenes, that works!

nacnudus:
The Notebook no longer works for me.

nacnudus:
The notebook works if you do the same for rlang as for httr.  Both of them have to be overridden. I'll comment on the gist too.

jobdiogenes:
> The notebook works if you do the same for rlang as for httr. Both of them have to be overridden. I'll comment on the gist too.

The notebook are working again. Although reassign is_interactive in rlang solves. 
I opt to use `options(rlang_interactive=TRUE)` which also solve. I hope that in future the use of options could be enough.

jennybc:
This conversation should be informed by https://github.com/r-lib/rlang/pull/1033.

It doesn't get at or solve the heart of the issue, but it's related.

PR #637, opened 2020-03-06
Update test-oauth.R

### LINKED ISSUE #636: tests/testtthat/test-oauth.R: The last test requires httpuv
If the package is not installed, the function won't throw an error.

michaelquinn32:
Fixes #636


hadley:
FWIW we usually assume that all suggested packages are installed when testing and we'll only add skips if they're a particularly arduous dependency.

michaelquinn32:
That's a Google thing. httpuv would be an "arduous" dependency for us.

Thanks!

PR #641, opened 2020-04-16
Add support for connections as body in POST and friends

s-u:
Allows `body` to be a connection. It uses `readfunction` (just like `form_file`) and chunked transfer encoding to allow `POST`s of non-seekable content where length is not known a priori (e.g. `pipe()` or others). It is implemented in `body_config` so works really for any method that requires body.


hadley:

PR #642, opened 2020-04-19
[docs] added RETRY() suggestion to api-packages vignette (fixes #638)

### LINKED ISSUE #638: Mention the RETRY function in the "Best practices for API packages" vignette?
The vignette has a paragraph starting with "For many APIs, the common approach is to retry API calls that return something in the 500 range." but it doesn't mention the [`RETRY()` function](https://httr.r-lib.org/reference/RETRY.html).

jameslamb:
Thanks for an awesome project! I use `httr` every day, and recommend it to anyone who will listen.

In this PR, I'd like to recommend an addition to the `api-packages` vignette. The vignette correctly mentions the importance of retrying failed requests and using exponential backoff w hen you do that, but it doesn't currently mention that `httr::RETRY()` already has an implementation of that and should be used. I'd like to propose an explicit call-out for `RETRY()` in the vignette.

For context, I'm working as part of a group at the Chicago R Collab conference this weekend, trying to improve the reliability of HTTP client packages on CRAN. We've found that many such clients use the single-shot `httr::<method>()` functions, and I suspect that that's because package authors aren't familiar with `RETRY()`. We've documented a lot more about our thinking (and the contributions we've made to other projects so far) at https://github.com/chircollab/chircollab20/issues/1.

Thanks for your time and consideration!


jameslamb:
I just rebased this to `master` to pick up things that have changed since I opened it.

jameslamb:
I just merged `master` into this to pick up more changes that have been added there

maelle:
This would fix #638

jameslamb:
This PR has been open for 20+ months with no interaction from maintainers, so I'm confident that there isn't interest in it. I'm going to close this.

PR #645, opened 2020-05-02
accept() takes 'type' vectors with length > 1

mtmorgan:
`Accept` headers can contain multiple types, e.g., `"application/json, text/plain"`, but `accept()` previously produced a warning (because `if (substr(type, 1, 1) == ".")` is non-scalar)  on vector arguments with length > 1


hadley:

PR #662, opened 2020-07-14
add a small snippet to ?timeout about further info on timeout

### LINKED ISSUE #658: Documentation doesn't mention defaults for timeout
I checked the CRAN manual and a default for `timeout` is not mentioned (maybe it's 0?)

I think a mention on both `?timeout` and `?GET` (perhaps the other REST methods' pages as well) would be helpful

MichaelChirico:
Closes #658 

I opted for a very minimal version for now. As Jenny pointed out, it can be quite complicated. Not sure what's worth spelling out in `httr` vs. just redirecting to `curl`.

Happy to expand as needed.

### DIFF
--- a/R/timeout.r
+++ b/R/timeout.r
@@ -3,7 +3,8 @@
 #' @param seconds number of seconds to wait for a response until giving up.
 #'   Can not be less than 1 ms.
 #' @family config
-#' @details This timeout is passed on to \code{\link[curl]{handle_setopt}}. See there and \code{\link[curl]{curl_options}} for more details.
+#' @details This timeout is passed on to [curl::handle_setopt()].
+#'   See there and [curl::curl_options()] for more details.
 #' @export
 #' @examples
 #' \dontrun{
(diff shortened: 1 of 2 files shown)


R/timeout.r
```diff
@@ -3,6 +3,7 @@
 #' @param seconds number of seconds to wait for a response until giving up.
 #'   Can not be less than 1 ms.
 #' @family config
+#' @details This timeout is passed on to \code{\link[curl]{handle_setopt}}. See there and \code{\link[curl]{curl_options}} for more details.
```
jennybc, line 6: Can you use markdown syntax and a line width similar to what you see elsewhere?

And maybe `document()`? If that causes more changes than you want to be responsible for, fair enough. Disregard in that case.

MichaelChirico: done, thanks

jennybc:
Thanks!

The CI failure doesn't seem related to this, so merging. `master` is failing for same reasons.

PR #670, opened 2020-08-14
Force numeric query params to not use scientific notation

gjuggler:
Potential fix for https://github.com/r-dbi/bigrquery/issues/395

I work on the All of Us Researcher Workbench project (https://github.com/all-of-us/workbench) where we guide users to fetch data from BigQuery via the `bigrquery` package. The issue above has been affecting our users, so we wanted to take a swing at a fix.

**Rationale**
The approach taken here is to treat numbers as special by pre-formatting them (and forcing non-scientific notation) before passing the string on to the `curl::curl_escape` function. (The `curl_escape` function uses as.character to convert its arguments to strings.)

This PR takes a stance that the prior behavior of auto-converting from 10000 to '1e+5' is not desired. If a user of this library wants a query param with scientific notation, they may convert the number to a scientific-formatted string before passing it to `httr`.

**Testing**
- devtools::check() runs successfully.
- I confirmed that the new `scientific notation is avoided` test fails if the code change is commented out.

**Notes**
The formatting change on the line above is arguably out-of-scope for this PR, but the [style guide](https://style.tidyverse.org/syntax.html#inline-statements) does suggest that return calls should be enclosed in braces.

### DIFF
--- a/R/url-query.r
+++ b/R/url-query.r
@@ -23,7 +23,12 @@ compose_query <- function(elements) {
   names <- curl::curl_escape(names(elements))
 
   encode <- function(x) {
-    if (inherits(x, "AsIs")) return(x)
+    if (inherits(x, "AsIs")) {
+      return(x)
+    }
+    if (is.numeric(x)) {
+      x <- format(x, scientific = FALSE)
+    }
     curl::curl_escape(x)
   }
   values <- vapply(elements, encode, character(1))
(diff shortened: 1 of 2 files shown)


gjuggler:
Note: we are pursuing a more limited fix to https://github.com/r-dbi/bigrquery/issues/395 within `bigrquery` itself, see https://github.com/r-dbi/bigrquery/pull/400. I'll leave this PR open in case it's eventually seen as important to encode this behavior within `httr`.

hadley:

PR #696, opened 2021-08-09
Tweak `content()` documentation

salim-b:


hadley:

PR #697, opened 2021-08-24
`request_perform()`: If the curl response does not return a valid url, use the request url

schloerke:
This change is involved in a wild goose chase of a bug. I hope the original problem is in https://github.com/jeroen/curl/issues/257.

Similar to https://github.com/rstudio/webdriver/pull/72, this PR should only be merge if no action comes from https://github.com/jeroen/curl/issues/257 . Marking this PR as a draft until a decision is made.

--------------------------

`resp` is somehow being returns as a stub of information from curl. This means the url is `""`, which cascades into many different errors. There are other unexpected pieces of information, such as no headers, but if a plausible url is provided then we can limp out of the function. 

If the `resp$url` equals `""` and it is not handled, then the `url_scheme` equals `NULL` which causes `is_http` to be `logical(0)`. This throws an error: `Error in if (is_http) {: argument is of length zero`. This error does not make sense when the original request url was valid.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # httr (development version)
 
+* When calling `request_perform()` and `curl` returns an incomplete response
+  object, use the original request url instead of an empty url. (#697)
+
 # httr 1.4.2
 
 * Fix failing test.
--- a/R/request.R
+++ b/R/request.R
@@ -156,7 +156,14 @@ request_perform <- function(req, handle, refresh = TRUE) {
     return(request_perform(req, handle, refresh = FALSE))
   }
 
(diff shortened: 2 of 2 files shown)


jeroen:
This cannot possibly be the right solution. I have never seen the "Failed initialization" errors, your server must be doing something very odd...

schloerke:
> This cannot possibly be the right solution 

Agreed. Closing this PR. 

> we better figure out the real problem

... phantomjs has been EOL for 4+ years? Hopefully shinytest2 (in dev) which uses chromote will naturally address this by removing phantomjs completely

PR #698, opened 2021-08-31
Make it clear httr uses different default than jsonlite

muschellij2:
`jsonlite::fromJSON` default uses `simplifyVector = TRUE`: https://github.com/jeroen/jsonlite/blob/master/R/fromJSON.R#L78
so making it so the docs noted that `httr::content` default was different.  Related to https://github.com/r-lib/httr/issues/85, but that's likely very dated (esp since it refers to `RJSONIO` vs. `jsonlite`)

@jeroen


hadley:

PR #701, opened 2021-09-10
Fixed typo in content_type() documentation

dlependorf:
There was a typo in the content_type() documentation, I fixed it and ran devtools::document() to carry it over to the corresponding .Rd file.


hadley:

PR #702, opened 2021-09-16
Fix argument value in return value docs

Bisaloo:
`as = "auto"` is not an authorised value. Based on the text, it should actually read `"parsed"`. This PR fixes this.


hadley:

PR #706, opened 2021-10-10
Add an escape control in `set_cookies`

womeimingzi11:
In some cases, we will get some cookies which is no need to make escape. With the missing of escape control, the twice cookies escape will generate wrong cookies.

To avoid the potential error, I have added `escape` argument in `httr:set_cookies`.


hadley:

PR #734, opened 2023-02-22
Make `is_interactive()` react to "rlang_interactive" option

jennybc:
This will allow me to make gargle's pseudo-OOB flow work in Google Colab, which would be really nice.

Relates to https://github.com/r-lib/gargle/pull/237, where I currently have `utils::assignInNamespace("is_interactive", rlang::is_interactive, ns = "httr")` in `.onLoad()`, but I can remove that if this PR gets merged and released. I don't think I could release gargle to CRAN with `assignInNamespace()`.

Relates to https://github.com/r-lib/gargle/issues/140

### DIFF
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -211,11 +211,9 @@ check_scope <- function(x) {
   paste(x, collapse = " ")
 }
 
-# Two reasons for wrapping base::interactive()
-# 1. Wrap in a non-primitive function, so it can be mocked for testing.
-# 2. gargle needs to access (pseudo-)OOB flow from Google Colab, so we need to
-#    be able to use the "rlang_interactive" option to signal that we are in
-#    an interactive environment.
+# gargle needs to access (pseudo-)OOB flow from Google Colab, so we need to
+# be able to use the "rlang_interactive" option to signal that we are in
+# an interactive environment. httr does not have an rlang dependency.
 is_interactive <- function() {
   getOption("rlang_interactive") %||% interactive()
 }
--- a/R/oauth-init.R
+++ b/R/oauth-init.R
@@ -211,8 +211,14 @@ check_scope <- function(x) {
   paste(x, collapse = " ")
 }
 
-# Wrap base::interactive in a non-primitive function so that the call can be mocked for testing
-is_interactive <- function() interactive()
+# Two reasons for wrapping base::interactive()
+# 1. Wrap in a non-primitive function, so it can be mocked for testing.
+# 2. gargle needs to access (pseudo-)OOB flow from Google Colab, so we need to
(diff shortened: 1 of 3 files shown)


R/oauth-init.R
```diff
@@ -211,8 +211,12 @@ check_scope <- function(x) {
...
 }
 
-# Wrap base::interactive in a non-primitive function so that the call can be mocked for testing
-is_interactive <- function() interactive()
+# gargle needs to access (pseudo-)OOB flow from Google Colab, so we need to
+# be able to use the "rlang_interactive" option to signal that we are in
+# an interactive environment. httr does not have an rlang dependency.
+is_interactive <- function() {
+  getOption("rlang_interactive") %||% interactive()
+}
```
jennybc, line 219: I can't use `rlang::is_interactive()` because httr does not depend on rlang at this point. And it doesn't seem worth bringing a whole standalone file in for this.

tests/testthat/test-oauth.R
```diff
@@ -111,13 +111,19 @@ test_that("oob must be a flag", {
 })
 
 test_that("can not use oob in non-interactive session", {
+  old <- options()
+  on.exit(options(old))
+  options(rlang_interactive = FALSE)
+
```
jennybc, line 117: This test would previously fail when run with `test_active_file()`, so I fixed that while I was here.

tests/testthat/test-oauth.R
```diff
@@ -111,13 +111,19 @@ test_that("oob must be a flag", {
...
 test_that("can not use custom oob value without enabling oob", {
   testthat::skip_if_not_installed("httpuv")
-  with_mock(
-    `httr:::is_interactive` = function() TRUE,
-    expect_error(check_oob(FALSE, "custom_value"), "custom oob value")
-  )
+
+  old <- options()
+  on.exit(options(old))
+  options(rlang_interactive = TRUE)
```
jennybc, line 126: Took the chance to get rid of the `with_mock()` call, even though httr doesn't use testthat 3e. httr also doesn't use withr.

hadley (approved):

jennybc:
Just to record this analysis somewhere: I have not replaced every call to `base::interactive()` with `is_interactive()`. Why not? Where else is `base::interactive()` called?

* `BROWSE()`: For my motivating case, which is Google Colab, `utils::browseURL()` does not work and we do indeed want to go down the non-`interactive()` branch.
* `should_cache()` (internal helper): gargle never hits this code, because caching is done in-house. But FWIW `utils::menu()` really does require a fully interactive session.
* Default args in `init_oauth1.0()`, `init_oauth2.0()`, `oauth_listener()`: these args are deprecated, so it doesn't matter.

PR #740, opened 2023-06-23
Tighten 'verb' description

MichaelChirico:
It wasn't immediately clear to me in `?RETRY` whether `RETRY(GET, ...)` would work. `is.name(GET)` is `TRUE` in R, after all.


hadley:

